PR #15, opened 2019-08-02
Post submission endpoint

armoj:
creation of the POST `/questionnaires/{questionnaireId}/submissions` endpoint. enables the consumer to kick off an application submission.


questionnaire/questionnaire.test.js
```diff
@@ -556,3 +584,361 @@ describe('/questionnaires/{questionnaireId}/sections/answers', () => {
         });
     });
 });
+describe('/questionnaires/{questionnaireId}/submissions', () => {
```
BarryPiccinni: As we have a number of top level 'describe' blocks now, it would be nice to wrap them all inside a 'describe' block so that tests can be executed in one go and output read at a glance. This also makes it is more consistent with the rest of our test files.

services/request/index.js
```diff
@@ -0,0 +1,55 @@
...
+        opts = merge(opts, options);
+        return got(options.url, opts);
+    }
+
+    function get(options) {
+        let opts = {
+            method: 'GET',
+            headers: {
+                accept: 'application/json',
+                'Content-Type': 'application/json'
```
BarryPiccinni: Should this be just `application/json` and not `application/vnd.api+json`

armoj: well caught! I'll add it in

armoj: This is being handled in the `message-bus` service file!

PR #20, opened 2019-08-05
Validate all answers

armoj:


BarryPiccinni:
@armoj Are there any tests to be added for this?

PR #21, opened 2019-08-07
Update get all questionnaire answers to work with real data

armoj:


questionnaire/questionnaire-service.js
```diff
@@ -234,6 +234,32 @@ function createQuestionnaireService(spec) {
         return response;
     }
 
+    async function getAnswers(questionnaireId) {
+        const result = await getQuestionnaire(questionnaireId);
+        const {questionnaire} = result.rows[0];
+
+        const resourceCollection = Object.keys(questionnaire.answers).reduce(
```
sinclairs: Can we do the inverse of this and map from progress? e.g.

```javascript
const resourceCollection = questionnaire.progress.map(sectionAnswersId => ({
    type: 'answers',
    id: sectionAnswersId,
    attributes: questionnaire.answers[sectionAnswersId]
}));
```

PR #30, opened 2019-08-13
Update express openapi validator

sinclairs:
We were using v0.53 of open api validator which was failing to properly validate querystring values against our openapi schema. Updating to the latest version (1.4.0) had changes in behaviour:

1. If "additionalProperties: false" is set, the new version removes any additional properties prior to validation. This gives false positives causing our tests to fail.
2. Unknown string formats in schemas now throw an error instead of being ignored

To address issue 1 I've forked the module and set "removeAdditional" to false.
To address issue 2 I've removed the unknown format "UUIDv4" from the questionnaireId param schema. It wasn't being used as part of the validation so safe to do so.

All tests are now passing again.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
         "express": "~4.16.0",
         "express-jwt": "^5.3.1",
         "express-jwt-authz": "^2.3.1",
-        "express-openapi-validator": "^0.53.3",
+        "express-openapi-validator": "git+https://github.com/webstacker/express-openapi-validator.git",
         "got": "^9.6.0",
         "helmet": "^3.15.0",
         "json-pointer": "^0.6.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -22,7 +22,7 @@
         "express": "~4.16.0",
         "express-jwt": "^5.3.1",
         "express-jwt-authz": "^2.3.1",
-        "express-openapi-validator": "^0.53.3",
+        "express-openapi-validator": "git+https://github.com/webstacker/express-openapi-validator.git",
```
armoj: Should this be forked from CICA?

sinclairs: I've got an open issue with the authors at the moment, hence the fork is on my account. Don't think it matters for the moment. Depending on their response we might need a permanent from here, otherwise we can ditch this and use their updated version from npm.

armoj: My thinking was that if the fix needed to be a permanent fork, then it would be be most fitting to have it in the CICA GitHub account. Use this at the moment, and we can change it if needed at a later date!

sinclairs: 👍

PR #35, opened 2019-08-21
Add pipe resolver

armoj:


sinclairs:
This has been superseded by #51. #51 extends this PR with support for the full https://tools.ietf.org/html/rfc6901 grammar, and escapes all special characters.

PR #45, opened 2019-09-03
update template

BarryPiccinni:


questionnaire/templates.js
```diff
@@ -1539,7 +1539,7 @@ module.exports = {
                         description:
                             'We may use this to contact you if we need to clarify something on your application form.',
                         maxLength: 20,
-                        pattern: '^(\\+|\\d)([\\d\\s\\(\\)\\+\\-\\#]){1,19}$',
+                        pattern: '^(\\+|\\d)([\\d\\s\\(\\)\\+\\-\\#]){7,19}$',
```
sinclairs: We can remove the capturing groups from this. Also the "\s"  can be replaced with " " as we only want spaces, not any whitespace e.g. "^[\\+\\d][\\d \\(\\)\\+\\-\\#]{7,19}$"

Just to confirm, were saying the minimum is 8 characters?

PR #59, opened 2019-09-24
Update query result check to use result.rowCount

armoj:
please review an merge this PR after #56, #57, #58 are merged


questionnaire/questionnaire-dal.js
```diff
@@ -30,6 +29,14 @@ function questionnaireDAL(spec) {
                 // 'UPDATE questionnaire SET questionnaire = jsonb_set(questionnaire, $1, $2, TRUE), modified = current_timestamp WHERE id = $3',
                 // [`{answers,${sectionId}}`, answers, questionnaireId]
             );
+            if (result.rowCount !== 1) {
```
sinclairs: can we change these to "=== 0"

PR #60, opened 2019-09-24
Update HSTS to be 365 days

armoj:


app.js
```diff
@@ -49,6 +49,11 @@ app.use(
         }
     })
 );
+app.use(
```
sinclairs: Can the helmet constructor/factory accept these arguments? Would allow us to remove some middleware calls

sinclairs: Can the helmet constructor/factory accept these arguments? Would allow us to remove some middleware calls

PR #71, opened 2019-10-17
Add template content updates

armoj:


armoj:
> As long as these changes have come via q-definitions? Looks good

yeah. q-definitions has been updated and these were generated from it.

PR #90, opened 2020-01-31
SMS

armoj:
Add SMS sending functionality.
* Add new 'confirmation method' screen.
* Add AJV validation for UK mobile phone numbers only
* minor JSON Pointer replacer logic update - allows pointers to potentially not resolve


services/replace-json-pointer/replace-json-pointer.test.js
```diff
@@ -47,16 +47,26 @@ describe('replace-json-pointer', () => {
...
-        }).toThrow('Invalid reference token: does-not-exist');
+    //     expect(() => {
+    //         replaceJsonPointer(content, {
+    //             foo: {
+    //                 bar: 'bar'
+    //             }
+    //         });
+    //     }).toThrow('Invalid reference token: does-not-exist');
+    // });
+
```
BarryPiccinni: Remove if not needed.

services/notify/index.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+process.env.NODE_TLS_REJECT_UNAUTHORIZED = 0;
```
BarryPiccinni: Should this be getting set here??

package.json
```diff
@@ -23,15 +23,19 @@
         "express-jwt": "^5.3.1",
         "express-jwt-authz": "^2.3.1",
         "express-openapi-validator": "^1.6.2",
+        "google-libphonenumber": "^3.2.6",
         "got": "^9.6.0",
-        "helmet": "^3.15.0",
+        "helmet": "^3.21.2",
         "json-pointer": "^0.6.0",
+        "libphonenumber-js": "^1.7.30",
```
sinclairs: Is this and google-libphonenumber required?

package.json
```diff
@@ -75,12 +79,12 @@
...
         "blocks": [
-            ".eslintignore",
             ".eslintrc.js",
-            ".lintstagedrc.js",
+            ".eslintignore",
             ".huskyrc.js",
-            ".prettierrc.js",
             "jest.config.js",
+            ".prettierrc.js",
+            ".lintstagedrc.js",
```
sinclairs: Just wondering if these were manually changed? The base config manages these

questionnaire/questionnaire-service.js
```diff
@@ -10,6 +10,8 @@ const templates = require('./templates');
 const createQuestionnaireDAL = require('./questionnaire-dal');
 const createMessageBusCaller = require('../services/message-bus');
 const replaceJsonPointers = require('../services/replace-json-pointer');
+const createNotifyService = require('../services/notify');
+const createValidatorService = require('../services/validator');
```
sinclairs: Can this be named to communicate what it's validating

questionnaire/questionnaire-service.js
```diff
@@ -23,6 +25,11 @@ function createQuestionnaireService(spec) {
 
     AjvErrors(ajv);
 
+    ajv.addFormat('x-mobilePhoneNumber', mobilePhoneNumber => {
+        const validator = createValidatorService();
```
sinclairs: Do we need a new instance per request?

package.json
armoj: removed

package.json
```diff
@@ -23,15 +23,19 @@
...
         "json-pointer": "^0.6.0",
+        "libphonenumber-js": "^1.7.30",
         "lodash.merge": "^4.6.2",
         "morgan": "^1.9.1",
+        "notifications-node-client": "^4.7.0",
         "pg": "^7.11.0",
         "pino": "^5.12.6",
         "pino-http": "^4.2.0",
         "q-router": "github:CriminalInjuriesCompensationAuthority/q-router#v2.4.0",
+        "request": "^2.88.0",
```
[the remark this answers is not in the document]
armoj: `request` used to be a "sub" dependency of `notification-node-client`. This is now not the case. It is needed as a dependency at this level. This is the official recommendation for the `notification-node-client`.

package.json
armoj: I don't remember changing these as I know they are managed elsewhere. I've changed them back to the order they were originally.

questionnaire/questionnaire-service.js
armoj: `ajv` is declared and initialised just above this within the same scope. This is the earliest the new format can be defined. Are you suggesting a bigger change here than moving this `addFormat` block?

questionnaire/questionnaire-service.js
```diff
@@ -23,6 +25,11 @@ function createQuestionnaireService(spec) {
 
     AjvErrors(ajv);
 
+    ajv.addFormat('x-mobilePhoneNumber', mobilePhoneNumber => {
+        const mobilePhoneNumberValidator = createMobilePhoneNumberValidatorService();
+        return mobilePhoneNumberValidator.isValidMobilePhoneNumber(mobilePhoneNumber);
+    });
+
```
armoj: @sinclairs `ajv` is defined just above this within the same scope. This is the earliest that I can define a new "format" for AJV. Is this what you were originally meaning or is there another issue you were talking about?

PR #107, opened 2020-04-21
Add custom AJV format module

armoj:
Abstracts out the custom AJV formats module that deals with UK mobile phone number validation


BarryPiccinni:
@armoj If this is now pulled from a module, can the function be removed from the services folder?

PR #120, opened 2020-06-08
Add the origin of the application

BarryPiccinni:
This commit adds the origin of the application to the postgres table.
This can be either 'TEL' to signify that the application was completed
through the CSC or 'WEB' to say it came from an external source.


questionnaire/questionnaire-service.js
```diff
@@ -552,6 +554,20 @@ function createQuestionnaireService(spec) {
         };
     }
 
+    async function getMeta(questionnaireId) {
```
armoj, line 557: can this be named `getQuestionnaireMetadata` just to be specific

PR #136, opened 2020-09-18
Add GET dataset endpoint

armoj:
Add `/questionnaires/{questionnaireId}/dataset` endpoint that returns a normalised set of answers and flattened list of answers.


armoj:
duplicate

PR #144, opened 2020-10-22
Update contact information

armoj:
* Make all contact information consistent
* content now in a partial

not to be merged in for the pilot. This branch is ahead of master and the changes apply to this cut of the code

### DIFF
--- a/questionnaire/templates.js
+++ b/questionnaire/templates.js
@@ -4,7 +4,7 @@ module.exports = {
     'sexual-assault': id => ({
         id,
         type: 'apply-for-compensation',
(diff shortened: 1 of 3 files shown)


questionnaire/templates.js
```diff
@@ -985,7 +985,7 @@ module.exports = {
                 properties: {
                     'you-need-to-ccontact-us': {
                         description:
```
BarryPiccinni, line 987: Can you fix the typo here anyway, just so it doesn't get lost in the aether?

BarryPiccinni (changes requested):

PR #150, opened 2020-11-03
Failed submission fix

armoj:
* Allows application with a submission status of `FAILED` to be able to be resubmitted.
* add tests

### DIFF
--- a/questionnaire/routes.js
+++ b/questionnaire/routes.js
@@ -147,8 +147,6 @@ router
         try {
             const {questionnaireId} = req.params;
             const questionnaireService = createQuestionnaireService({logger: req.log});
-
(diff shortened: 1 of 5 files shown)


questionnaire/routes.js
```diff
@@ -190,24 +196,19 @@ router
 
             // if the summary section ID is in the progress array, then that means
             // the questionnaire is submittable.
```
BarryPiccinni, line 198: This comment is still valid and meaningful but is now in the wrong place. Could it be moved to around line 174

BarryPiccinni (changes requested):

BarryPiccinni (approved):

PR #157, opened 2020-12-14
Add confirm incident type

armoj:
* remove redundant triage question schemas
* add `incident type` schema
* update routing to cater for sexual, and non-sexual assault selection from the `incident type` schema


sinclairs (approved):
Julie and Barry K have signed-off on these changes

PR #166, opened 2021-03-25
Trigger notification on event

armoj:
* Move `sendConfirmationNotification` invocation to the system endpoint handler.
* Update tests

### DIFF
--- a/questionnaire/questionnaire-service.js
+++ b/questionnaire/questionnaire-service.js
@@ -190,16 +190,6 @@ function createQuestionnaireService({
     async function getSubmissionResponseData(questionnaireId, isPostRequest = false) {
         let submissionStatus = await getQuestionnaireSubmissionStatus(questionnaireId);
 
-        let submitted = false;
-        let status = 'NOT_STARTED';
-        let caseReferenceNumber = null;
-
-        // if it is already completed, just get the Case Reference Number then
-        // let it fall through to the return.
-        if (submissionStatus === 'COMPLETED') {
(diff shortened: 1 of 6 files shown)


questionnaire/questionnaire-service.js
```diff
@@ -215,7 +215,6 @@ function createQuestionnaireService({
             // if the caseReferenceNumber exists, then the submission is COMPLETE.
             if (caseReferenceNumber) {
                 await updateQuestionnaireSubmissionStatus(questionnaireId, 'COMPLETED');
```
sinclairs, line 217: Should this not treated the same as sending a notification? e.g. they'll get a notification eventually, however their application will still show "IN_PROGRESS"

questionnaire/routes.js
```diff
@@ -83,6 +83,7 @@ router
                 'system',
                 answers
             );
+            await questionnaireService.sendConfirmationNotification(req.params.questionnaireId);
```
sinclairs, line 86: for a tiny performance bump... does this need to be awaited?

sinclairs (changes requested):

questionnaire/routes.js
armoj: nope it doesn't! wasn't originally. don't know why I added it in. Good catch

questionnaire/questionnaire-service.js
armoj: If you are basically asking: "should we move the updating of the submission_status in to the `system answers` route handler?" then I did think about that yeah. Maybe it should be that way!

BarryPiccinni (approved):

PR #180, opened 2021-05-11
Bump base-config

armoj:
Bump base-config to `v0.10.0`


sinclairs (changes requested):
@armoj This is failing at the audit step due to not have the bumped `json-pointer` module from master. Can you grab the master changes and I can approve this? Looks good otherwise 👍

sinclairs (approved):

PR #183, opened 2021-05-13
Add resubmission of failed endpoint

armoj:
* Add `"/questionnaires/submissions/resubmit-failed"` endpoint.
* Add `getQuestionnaireIdsBySubmissionStatus` DAL function.
* Add `ResourceConflicts` error to cater for submission resource errors that are thrown.
* Rename `getSubmissionResponseData` to `getSubmissionResource`.
* Update tests to actually test the submission functionality directly.
* Update Jest config to allow tests through.

Manually tested using Postman to hit the new endpoint. Emulated `FAILED` submission via row updates using PG Admin.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -29,10 +29,10 @@ config.coverageThreshold = {
     './questionnaire/questionnaire-dal.js': {
         branches: 0,
         functions: 0,
-        lines: 9,
-        statements: 9
+        lines: 7,
+        statements: 7
     },
-    './*/!(message-bus|notify|request)/**/*.js': {
+    './*/!(message-bus|notify|request|error-handler)/**/*.js': {
         statements: 54,
         branches: 60,
         functions: 50,
@@ -55,6 +55,12 @@ config.coverageThreshold = {
         branches: 60,
         functions: 0,
         lines: 23
+    },
+    './*/error-handler/**/*.js': {
+        statements: 45,
+        branches: 60,
+        functions: 40,
+        lines: 47
     }
 };
 
--- a/middleware/error-handler/index.js
+++ b/middleware/error-handler/index.js
@@ -141,7 +141,7 @@ module.exports = async (err, req, res, next) => {
             submissions: errorInfo.submissions
         };
 
-        return res.status(200).json(error);
+        return res.status(422).json(error);
     }
 
     if (err.name === 'UpdateNotSuccessful') {
@@ -197,7 +197,7 @@ module.exports = async (err, req, res, next) => {
     if (err.statusCode === 409) {
         error.errors.push({
             status: 409,
-            title: '409 Resource Conflict',
+            title: '409 Conflict',
             detail: err.message
         });
 
@@ -209,7 +209,7 @@ module.exports = async (err, req, res, next) => {
         const errors = errorInfo.errors.map(errorObj => {
             return {
                 status: 409,
-                title: '409 Resource Conflict',
+                title: '409 Conflict',
                 detail: errorObj.message
             };
         });
@@ -219,7 +219,7 @@ module.exports = async (err, req, res, next) => {
             submissions: errorInfo.submissions
         };
 
-        return res.status(200).json(error);
+        return res.status(409).json(error);
     }
 
     if (err.name === 'UnauthorizedError') {
--- a/middleware/error-handler/index.js
+++ b/middleware/error-handler/index.js
@@ -197,11 +197,29 @@ module.exports = async (err, req, res, next) => {
     if (err.statusCode === 409) {
         error.errors.push({
             status: 409,
-            title: '409 Conflict',
+            title: '409 Resource Conflict',
             detail: err.message
         });
 
         return res.status(409).json(error);
+    }
+
+    if (err.name === 'ResourceConflicts') {
+        const errorInfo = VError.info(err);
+        const errors = errorInfo.errors.map(errorObj => {
+            return {
+                status: 409,
+                title: '409 Resource Conflict',
+                detail: errorObj.message
+            };
+        });
+
+        error.errors.push(...errors);
+        error.meta = {
+            submissions: errorInfo.submissions
+        };
+
+        return res.status(200).json(error);
     }
 
     if (err.name === 'UnauthorizedError') {
--- a/questionnaire/questionnaire-dal.js
+++ b/questionnaire/questionnaire-dal.js
@@ -125,15 +125,11 @@ function questionnaireDAL(spec) {
             result = await db.query('SELECT id FROM questionnaire WHERE submission_status = $1', [
                 submissionStatus
             ]);
-
-            if (result.rowCount === 0) {
-                result = [];
-            }
         } catch (err) {
             throw err;
         }
 
-        return result.rows.map(x => x.id);
+        return result.rowCount ? result.rows.map(x => x.id) : [];
     }
 
     return Object.freeze({
--- a/questionnaire/questionnaire-dal.js
+++ b/questionnaire/questionnaire-dal.js
@@ -118,12 +118,36 @@ function questionnaireDAL(spec) {
         return result;
     }
 
+    async function getQuestionnaireIdsBySubmissionStatus(submissionStatus) {
+        let result;
+
+        try {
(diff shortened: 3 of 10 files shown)


jest.config.js
```diff
@@ -29,10 +29,10 @@ config.coverageThreshold = {
     './questionnaire/questionnaire-dal.js': {
         branches: 0,
         functions: 0,
-        lines: 9,
-        statements: 9
+        lines: 7,
+        statements: 7
```
sinclairs, line 33: These thresholds are really poor as it is. We shouldn't lower them any further. Can the test coverage be improved to at least maintain the level?

jest.config.js
```diff
@@ -55,6 +55,12 @@ config.coverageThreshold = {
         branches: 60,
         functions: 0,
         lines: 23
+    },
+    './*/error-handler/**/*.js': {
+        statements: 45,
+        branches: 60,
+        functions: 40,
+        lines: 47
```
sinclairs, line 63: These thresholds aren't great as it is. We shouldn't lower them any further. Can the test coverage be improved to at least maintain the level?

middleware/error-handler/index.js
```diff
@@ -197,13 +197,31 @@ module.exports = async (err, req, res, next) => {
     if (err.statusCode === 409) {
         error.errors.push({
             status: 409,
-            title: '409 Conflict',
+            title: '409 Resource Conflict',
```
sinclairs, line 200: For consistency with other status codes, can this just remain `409 Conflict`?

questionnaire/questionnaire-dal.js
```diff
@@ -118,12 +118,36 @@ function questionnaireDAL(spec) {
...
+    async function getQuestionnaireIdsBySubmissionStatus(submissionStatus) {
+        let result;
+
+        try {
+            result = await db.query('SELECT id FROM questionnaire WHERE submission_status = $1', [
+                submissionStatus
+            ]);
+
+            if (result.rowCount === 0) {
+                throw new VError(
```
sinclairs, line 130: I'd expect something that selects by status to return 0 or more things, should this be an exception?

questionnaire/questionnaire-service.js
```diff
@@ -209,20 +209,18 @@ function createQuestionnaireService({
...
-            data: {
-                id: questionnaireId,
-                type: 'submissions',
-                attributes: {
-                    questionnaireId,
-                    submitted,
-                    status,
-                    caseReferenceNumber
-                }
+        const resource = {
```
sinclairs, line 212: Does the top level `data` attribute need to move?

questionnaire/questionnaire-service.js
```diff
@@ -588,18 +586,131 @@ function createQuestionnaireService({
         };
     }
 
+    async function createSubmission(questionnaireId) {
+        // 1) get questionnaire instance.
+        const questionnaire = await getQuestionnaire(questionnaireId);
+        if (!questionnaire) {
```
sinclairs, line 592: `getQuestionnaire` ends up throwing already, not sure this will be hit

sinclairs (changes requested):
This is much bigger than I'd expected. Be good to talk it though.

questionnaire/questionnaire-service.js
```diff
@@ -588,18 +586,131 @@ function createQuestionnaireService({
...
+        // means that the submission resource has been previously created.
+        // also skip over this for failed application so they can be resubmitted.
+        if (!['NOT_STARTED', 'FAILED'].includes(submissionStatus)) {
+            throw new VError(
+                {
+                    name: 'ResourceConflict'
+                },
+                `Submission resource with ID "${questionnaireId}" already exists`
+            );
+        }
```
BarryPiccinni, line 628: Is this logic definitely correct/necessary? We don't want it to try to submit anything that's not ready to submit, but these cases shouldn't be returned by the query anyway. Also, should we be limiting the type of status we can resubmit? I'm sure we discussed this, but while it **_shouldn't_** be able to get stuck `IN_PROGRESS` anymore, but what if it does? Was it agreed we ignore this?

questionnaire/questionnaire-service.js
```diff
@@ -588,18 +586,131 @@ function createQuestionnaireService({
...
+
+        // check all answers are correct.
+        await validateAllAnswers(questionnaireId);
```
BarryPiccinni, line 631: Is this check necessary? What will it do if the answers aren't correct?

questionnaire/questionnaire-service.js
```diff
@@ -588,18 +586,131 @@ function createQuestionnaireService({
...
+
+    async function postSubmissions(submissionStatus) {
+        const status = submissionStatus.toUpperCase();
+        let questionnaireIds;
+        try {
+            questionnaireIds = await db.getQuestionnaireIdsBySubmissionStatus(status);
+        } catch (err) {
+            // no IDs means there is nothing to submit, so return early.
+            return [];
+        }
```
BarryPiccinni, line 661: If it's already thrown in the `getQuestionnaireIdsBySubmissionStatus` function, I don't know that this will ever catch anything? If it does though, an empty array doesn't seem very helpful.

questionnaire/routes.js
```diff
@@ -147,64 +149,27 @@ router
...
+router
+    .route('/submissions/resubmit-failed')
+    .post(permissions('update:questionnaires'), async (req, res, next) => {
+        try {
+            const questionnaireService = createQuestionnaireService({logger: req.log});
+            const resource = await questionnaireService.postSubmissions('failed');
 
-            res.status(201).json(response);
+            res.status(201).json({
+                data: resource
```
BarryPiccinni, line 171: Why has this changed?

questionnaire/routes.js
```diff
@@ -136,9 +136,11 @@ router
                 throw err;
             }
 
-            const response = await questionnaireService.getSubmissionResponseData(questionnaireId);
+            const resource = await questionnaireService.getSubmissionResource(questionnaireId);
 
-            res.status(200).json(response);
+            res.status(200).json({
+                data: resource
```
BarryPiccinni, line 142: Why has this changed?

questionnaire/routes.js
```diff
@@ -147,64 +149,27 @@ router
...
+        }
+    });
 
-            questionnaireService.createAnswers(questionnaireId, questionnaire.routes.summary, {});
+router
+    .route('/submissions/resubmit-failed')
+    .post(permissions('update:questionnaires'), async (req, res, next) => {
+        try {
+            const questionnaireService = createQuestionnaireService({logger: req.log});
+            const resource = await questionnaireService.postSubmissions('failed');
```
BarryPiccinni, line 168: Are we only resubmitting failed applications?

BarryPiccinni (changes requested):
Maybe not something to address in this PR, but I see the coverage thresholds have been lowered. The test coverage for this repo is not great, particularly the DB stuff, some of which is completely untested.

questionnaire/questionnaire-service.js
armoj: this pattern conforms to the rest of the resources return from their respective functions. I want to return the resource and then shape the request (with the data attribute) later

jest.config.js
armoj: They are terrible and have been for some time - 9% was terrible in the first place. However I do not want to add more stuff to this PR.

jest.config.js
armoj: see comment above

questionnaire/questionnaire-service.js
armoj: removed. Hangover from previous iteration

sinclairs:
@armoj @BarryPiccinni Before more effort is put in to this PR, can we catch up on Monday? It'd be good to understand a few things:

1.  There are going to be conflicts with the contextualised work in regards to the OpenAPI changes. You never mentioned you were working on that too. We'll need to agree on how we're getting these changes in.
2. Given our conversation the other day around the current submission process probably not being the way to go moving forward, is this level of refactoring required?
3. If the intention is to hit this with an automated job, are the response codes etc necessary? I'm not clear on how they'll be used?
4. What are the blockers around maintaining the existing test thresholds?

PR #190, opened 2021-05-18
Update section answers resource schema

sinclairs:
Updates the section answers resource to use the common schemas.

All tests passing

### DIFF
--- a/openapi/src/json-schemas/api/resources/answer-resource.json
+++ b/openapi/src/json-schemas/api/resources/answer-resource.json
@@ -1,0 +1,14 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "Answer resource",
+    "allOf": [
+        {"$ref": "../json-api-base-resource.json"},
+        {
+            "properties": {
+                "type": {
+                    "const": "answers"
+                }
+            }
+        }
+    ]
+}
--- a/openapi/openapi.json
+++ b/openapi/openapi.json
@@ -663,35 +663,95 @@
                         "content": {
                             "application/vnd.api+json": {
                                 "schema": {
-                                    "type": "object",
-                                    "additionalProperties": false,
-                                    "required": ["data"],
-                                    "properties": {
-                                        "data": {
+                                    "title": "Single answers document",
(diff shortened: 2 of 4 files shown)


openapi/src/json-schemas/api/resources/answer-resource.json
```diff
@@ -0,0 +1,14 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "Answer resource",
+    "allOf": [
+        {"$ref": "../json-api-base-resource.json"},
+        {
+            "properties": {
+                "type": {
+                    "const": "answers"
+                }
```
armoj, line 10: I know that `attributes` is defined in `json-api-base-resource.json`, but should this have a specific attributes definition?

armoj (changes requested):

openapi/src/json-schemas/api/resources/answer-resource.json
sinclairs: It could, I'm not sure of the value, certainly with OpenAPI. It wouldn't be able to check that the keys are valid question ids as there are is no pattern support, and the value of the keys can be any JSON ,which is what the "{}" schema allows

sinclairs: Any thoughts on what could be added?

armoj: I can see that the value can be any JSON. was just wondering if it was worth tightening it at all at this point

armoj: I was thinking that there was a discernible shape that would not require `patternProperties` to be defined here. But that is exactly what it is. Had arrays on the brain.

sinclairs: That's what I mean, I'm not too sure on what else could be added, any thoughts?

armoj: no! I was thinking that the answers was an array for whatever reason. ignore me

armoj (approved):

PR #191, opened 2021-05-18
Update all section answers resource schema

sinclairs:
Updates the all section answers resource to use the common schemas.

All tests passing

### DIFF
--- a/openapi/src/json-schemas/api/resources/answer-resource.json
+++ b/openapi/src/json-schemas/api/resources/answer-resource.json
@@ -1,0 +1,14 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "Answer resource",
+    "allOf": [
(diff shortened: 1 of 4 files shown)


openapi/src/json-schemas/api/resources/answer-resource.json
```diff
@@ -0,0 +1,14 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "Answer resource",
+    "allOf": [
+        {"$ref": "../json-api-base-resource.json"},
+        {
+            "properties": {
+                "type": {
+                    "const": "answers"
+                }
```
armoj, line 10: should we define a more specific `attributes` here and not just "inherit" the generic one?

armoj (changes requested):

armoj (approved):

PR #192, opened 2021-05-18
Update dataset resource schema

sinclairs:
Updates the dataset resource to use the common schemas.

All tests passing

### DIFF
--- a/openapi/src/json-schemas/api/resources/dataset-resource.json
+++ b/openapi/src/json-schemas/api/resources/dataset-resource.json
@@ -1,0 +1,18 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "Dataset resource",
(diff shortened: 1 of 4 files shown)


openapi/src/json-schemas/api/resources/dataset-resource.json
```diff
@@ -0,0 +1,18 @@
...
+        {"$ref": "../json-api-base-resource.json"},
+        {
+            "properties": {
+                "type": {
+                    "const": "dataset"
+                },
+                "id": {
+                    "type": "string",
+                    "const": "0"
+                }
```
armoj, line 14: should we define a more specific `attributes` here and not just "inherit" the generic one?

armoj (changes requested):

armoj (approved):

PR #199, opened 2021-05-20
Add resubmission function

armoj:
* Add `/api/v1/submissions/resubmit-failed` endpoint.
* Add `getQuestionnaireIdsBySubmissionStatus` DAL function.
* Add tests

Tests pass.

Manually tested using Postman to hit the new endpoint. Emulated `FAILED` submission via row updates using PG Admin.

### DIFF
--- a/questionnaire/submissions-routes.js
+++ b/questionnaire/submissions-routes.js
@@ -1,0 +1,26 @@
+'use strict';
+
+const express = require('express');
+const validateJWT = require('express-jwt');
+
+const createSubmissionsService = require('./submissions-service');
+const permissions = require('../middleware/route-permissions');
+
+const router = express.Router();
+
+router.use(validateJWT({secret: process.env.DCS_JWT_SECRET}));
+
+router
+    .route('/resubmit-failed')
+    .post(permissions('update:questionnaires'), async (req, res, next) => {
+        let response;
+        try {
+            const submissionsService = createSubmissionsService({logger: req.log});
+            response = await submissionsService.postFailedSubmissions();
+        } catch (err) {
+            next(err);
+        }
+        res.status(200).json(response);
+    });
+
+module.exports = router;
--- a/questionnaire/submissions-service.js
+++ b/questionnaire/submissions-service.js
@@ -1,0 +1,41 @@
+'use strict';
+
+const createMessageBusCaller = require('../services/message-bus');
+const createQuestionnaireDAL = require('./questionnaire-dal');
+
+function createSubmissionService({logger}) {
+    const db = createQuestionnaireDAL({logger});
+
+    async function getQuestionnaireIdsBySubmissionStatus(status) {
+        return db.getQuestionnaireIdsBySubmissionStatus(status);
+    }
+
+    async function updateQuestionnaireSubmissionStatus(questionnaireId, submissionStatus) {
+        return db.updateQuestionnaireSubmissionStatus(questionnaireId, submissionStatus);
+    }
+
+    async function postFailedSubmissions() {
+        const questionnaireIds = await getQuestionnaireIdsBySubmissionStatus('FAILED');
+        const messageBusCaller = createMessageBusCaller({logger});
+        const promises = questionnaireIds.map(async id => {
+            await updateQuestionnaireSubmissionStatus(id, 'IN_PROGRESS');
+            try {
+                await messageBusCaller.post('SubmissionQueue', {
+                    applicationId: id
+                });
+                return {id, resubmitted: true};
+            } catch (err) {
+                await updateQuestionnaireSubmissionStatus(id, 'FAILED');
+                return {id, resubmitted: false};
+            }
+        });
+        const results = await Promise.all(promises);
+        return results;
+    }
+
+    return Object.freeze({
+        postFailedSubmissions
+    });
+}
+
+module.exports = createSubmissionService;
--- a/questionnaire/submissions.test.js
+++ b/questionnaire/submissions.test.js
@@ -40,10 +40,22 @@ jest.doMock('../services/slack', () => {
     }));
 });
 
+const messageBusPost = jest.fn(async (queueName, payload) => {
+    if (payload.applicationId === '4ddb0208-f7da-4237-a244-34e7e58d2ddf') {
+        throw new Error('some error');
+    }
+    return {body: 'Message sent'};
+});
+
+jest.doMock('../services/message-bus/index.js', () => {
+    return jest.fn(() => ({
+        post: messageBusPost
+    }));
+});
+
+const updateQuestionnaireSubmissionStatus = jest.fn();
 jest.doMock('./questionnaire-dal.js', () =>
-    // return a modified factory function, that returns an object with a method, that returns a valid created response
     jest.fn(() => ({
-        // createQuestionnaire: () => createQuestionnaireResponse,
         getQuestionnaire: questionnaireId => {
             if (
                 questionnaireId === '285cb104-0c15-4a9c-9840-cb1007f098fb' || // not started.
@@ -104,16 +116,14 @@ jest.doMock('./questionnaire-dal.js', () =>
                 `Questionnaire "${questionnaireId}" not found`
             );
         },
-        updateQuestionnaireSubmissionStatus: () => undefined,
-        retrieveCaseReferenceNumber: () => undefined
+        updateQuestionnaireSubmissionStatus,
+        retrieveCaseReferenceNumber: () => undefined,
+        getQuestionnaireIdsBySubmissionStatus: () => [
+            '4ddb0208-f7da-4237-a244-34e7e58d2ddf',
+            '93259f6d-1826-4e97-ba39-53b4e232dd81'
+        ]
     }))
 );
-
-// jest.doMock('../services/message-bus/index.js', () =>
-//     jest.fn(() => ({
-//         post: () => postSubmissionQueueResponse
-//     }))
-// );
 
 const app = require('../app');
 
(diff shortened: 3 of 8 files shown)


questionnaire/questionnaire-service.js
```diff
@@ -588,6 +588,27 @@ function createQuestionnaireService({
         };
     }
 
+    async function getQuestionnaireIdsBySubmissionStatus(status) {
+        return db.getQuestionnaireIdsBySubmissionStatus(status);
+    }
+
+    async function postSubmissionsByStatus(status) {
```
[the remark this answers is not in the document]
armoj: This is the reason I wanted to use the other functionality we chatted about. What I could do here is update the submission to IN_PROGRESS at this point and it would avoid that issue.

questionnaire/submissions.js
```diff
@@ -0,0 +1,25 @@
...
+const router = express.Router();
+
+router.use(validateJWT({secret: process.env.DCS_JWT_SECRET}));
+
+router
+    .route('/resubmit-failed')
+    .post(permissions('update:questionnaires'), async (req, res, next) => {
+        try {
+            const questionnaireService = createQuestionnaireService({logger: req.log});
+            const response = await questionnaireService.postSubmissionsByStatus('failed');
```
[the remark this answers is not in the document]
armoj: only if we are in favour of not caring about the response

questionnaire/submissions-service.js
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    async function postSubmissionsByStatus(status) {
+        const submissionStatus = status.toUpperCase();
+        const questionnaireIds = await getQuestionnaireIdsBySubmissionStatus(submissionStatus);
+        let response = [];
+        if (questionnaireIds.length) {
+            const messageBusCaller = createMessageBusCaller({logger});
+            const promises = questionnaireIds.map(async id => {
+                await updateQuestionnaireSubmissionStatus(id, 'IN_PROGRESS');
```
armoj, line 24: should this be await? I am thinking it should. having this be set to IN_PROGRESS first will avoid it being picked up by a subsequent request.

sinclairs: It looks like this could get stuck "IN_PROGRESS" should the message bus call fail

sinclairs (changes requested):

questionnaire/submissions-service.js
BarryPiccinni: I'm trying to follow the flow for this, if  the application service was down, I understand they would queue up and be processed eventually, but if the message bus was down would these promises not just return an error, leaving these cases `IN_PROGRESS` and unable to get picked up by a later job?

questionnaire/submissions-routes.js
```diff
@@ -0,0 +1,24 @@
...
+const router = express.Router();
+
+router.use(validateJWT({secret: process.env.DCS_JWT_SECRET}));
+
+router
+    .route('/resubmit-failed')
+    .post(permissions('update:questionnaires'), async (req, res, next) => {
+        try {
+            const submissionsService = createSubmissionsService({logger: req.log});
+            submissionsService.postSubmissionsByStatus('failed');
```
BarryPiccinni, line 18: Given that we now have a distinct submission function and a resubmission function and we shouldn't be able to fall into any unwanted/unexpected statuses, could we not just lock this to only find failed applications? Or are you thinking this still needs to be reusable for other statuses?

BarryPiccinni (changes requested):
As well as getting one failing test locally, I'm also getting quite a number of threshold violations that wont pass the CI pipeline test.

questionnaire/submissions-service.js
armoj: are you basically saying that we probably need a try/catch safety net to avoid that situation? Something like the below.

````
const promises = questionnaireIds.map(async id => {
                await updateQuestionnaireSubmissionStatus(id, 'IN_PROGRESS');
                let promise;
                try {
                    promise = messageBusCaller.post('SubmissionQueue', {
                        applicationId: id
                    });
                } catch (err) {
                    await updateQuestionnaireSubmissionStatus(id, 'FAILED');
                }
                return promise;
            });
````

File: questionnaire/submissions.test.js [hunk 5]
```diff
@@ -40,10 +40,34 @@ jest.doMock('../services/slack', () => {
...
+        ].includes(payload.applicationId)
+    ) {
+        throw new Error('some error');
+    }
+});
+jest.doMock('../services/message-bus/index.js', () => {
+    return jest.fn(() => ({
+        post: messageBusPost
+    }));
+});
```
sinclairs, line 59: This doesn't match the message bus caller behaviour

questionnaire/submissions.test.js
```diff
@@ -472,3 +496,41 @@ describe('Questionnaire submissions', () => {
...
+                const submissionsService = createSubmissionsService({logger: jest.fn()});
+                await submissionsService.postFailedSubmissions();
+                expect(questionnaireDAL.updateQuestionnaireSubmissionStatus).toHaveBeenCalledWith(
+                    '4ddb0208-f7da-4237-a244-34e7e58d2ddf',
+                    'FAILED'
+                );
+            });
+        });
+    });
+});
```
sinclairs, line 536: This test isn't accurate due to the invalid mock

questionnaire/submissions-service.js
```diff
@@ -0,0 +1,42 @@
...
+        const messageBusCaller = createMessageBusCaller({logger});
+        const promises = questionnaireIds.map(async id => {
+            await updateQuestionnaireSubmissionStatus(id, 'IN_PROGRESS');
+            let promise;
+            try {
+                promise = messageBusCaller.post('SubmissionQueue', {
+                    applicationId: id
+                });
+            } catch (err) {
+                await updateQuestionnaireSubmissionStatus(id, 'FAILED');
```
sinclairs, line 29: This won't be hit

questionnaire/submissions-service.js
```diff
@@ -0,0 +1,42 @@
...
+
+    async function updateQuestionnaireSubmissionStatus(questionnaireId, submissionStatus) {
+        return db.updateQuestionnaireSubmissionStatus(questionnaireId, submissionStatus);
+    }
+
+    async function postFailedSubmissions() {
+        const questionnaireIds = await getQuestionnaireIdsBySubmissionStatus('FAILED');
+        let response = [];
+        const messageBusCaller = createMessageBusCaller({logger});
+        const promises = questionnaireIds.map(async id => {
```
sinclairs, line 21: What's the intent for these promises vs fire and forget?

sinclairs (changes requested):
Appears to be a fundamental issue with the logic. I can review the rest when the comments have been addressed.

questionnaire/submissions-service.js
armoj: so that it is somewhat testable and so the coverage is not 0%

questionnaire/submissions-service.js
armoj: coverage says it is

questionnaire/submissions.test.js
armoj: again.. rubbish tests for a quick and dirty endpoint. this `throw` is actually emulating the `got` module throwing. as far as the submissions-service is concerned, the behaviour is identical. Not 100% ideal... however let's get something out. it effectively does exactly as expected.

sinclairs: Either I'm missing something? or this isn't doing what you think it is. `got` is asynchronous and therefore so is the message bus caller. This mock isn't asynchronous. I'm happy to get something out, but it needs to work as expected and I don't think this does.

questionnaire/submissions-service.js
sinclairs: I assume that's because the test is wrong? https://github.com/CriminalInjuriesCompensationAuthority/data-capture-service/pull/199#discussion_r644903934

questionnaire/submissions.test.js
armoj: this is just to force an error when a specific id is used. i.e. emulating a failure in a specific scenario. I will adjust it to be async as that was an error

sinclairs (changes requested):
Probs best talking through the rest of this review. I'll see if you're available on slack.

questionnaire/submissions.test.js
```diff
@@ -472,3 +489,46 @@ describe('Questionnaire submissions', () => {
...
+        });
+    });
+    describe('Message Bus NOT operational', () => {
+        it('should set a submission to FAILED', async () => {
+            // eslint-disable-next-line global-require
+            const createQuestionnaireDAL = require('./questionnaire-dal');
+            const questionnaireDAL = createQuestionnaireDAL({logger: jest.fn()});
+            const createSubmissionsService = jest.requireActual('./submissions-service');
+            const submissionsService = createSubmissionsService({logger: jest.fn()});
+            await submissionsService.postFailedSubmissions();
```
sinclairs, line 519: This test isn't independent. These lines could be removed and it would still pass due to the previous test

questionnaire/submissions.test.js
```diff
@@ -472,3 +489,46 @@ describe('Questionnaire submissions', () => {
...
+        });
+    });
+    it('should resubmit failed submissions', async () => {
+        const response = await request(app)
+            .post('/api/v1/submissions/resubmit-failed')
+            .set('Authorization', `Bearer ${tokens['update:questionnaires']}`)
+            .set('Content-Type', 'application/vnd.api+json')
+            .send({});
+        expect(response.body).toEqual({status: 'ok'});
+    });
```
sinclairs, line 533: This doesn't appear to assert "should resubmit failed submissions". Given you're returning from `postFailedSubmissions`, why not return that in the response and check that it's as expected?

sinclairs (changes requested):

questionnaire/submissions.test.js
armoj: that is what I originally had and that was said to be not a good approach. So I removed a "useful" response.

armoj: I can add it back in...

questionnaire/submissions.test.js
```diff
@@ -472,3 +489,46 @@ describe('Questionnaire submissions', () => {
...
+            expect(questionnaireDAL.updateQuestionnaireSubmissionStatus).toHaveBeenCalledWith(
+                '93259f6d-1826-4e97-ba39-53b4e232dd81',
+                'IN_PROGRESS'
+            );
+            expect(questionnaireDAL.updateQuestionnaireSubmissionStatus).not.toHaveBeenCalledWith(
+                '93259f6d-1826-4e97-ba39-53b4e232dd81',
+                'FAILED'
+            );
+        });
+    });
```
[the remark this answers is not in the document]
armoj: This test is to ensure that [this](https://github.com/CriminalInjuriesCompensationAuthority/data-capture-service/blob/add-resubmission-function/questionnaire/submissions-service.js#L22) line was hit, BUT not [this](https://github.com/CriminalInjuriesCompensationAuthority/data-capture-service/blob/add-resubmission-function/questionnaire/submissions-service.js#L29) line

questionnaire/submissions-routes.js
```diff
@@ -0,0 +1,27 @@
...
+
+router.use(validateJWT({secret: process.env.DCS_JWT_SECRET}));
+
+router
+    .route('/resubmit-failed')
+    .post(permissions('update:questionnaires'), async (req, res, next) => {
+        let response;
+        try {
+            const submissionsService = createSubmissionsService({logger: req.log});
+            // call it and forget about it. we don't need a return value.
```
sinclairs, line 19: Can you remove this as it no longer applies

sinclairs (changes requested):
Just a comment to remove.

Given the changes do you want to check on UAT before approval?

armoj:
> Just a comment to remove.
> 
> Given the changes do you want to check on UAT before approval?

sinclairs (approved):
Mind and check that this assumption is correct before deploying: https://github.com/CriminalInjuriesCompensationAuthority/data-capture-service/blob/bdf1c6fd6c83c7a4255a23ab6dd74d10273678a3/questionnaire/submissions-service.js#L23

PR #143, opened 2020-10-21
Abstract from dcs

armoj:
* abstract out the application template fully in to its own module
* remove redundant dependencies
* renamed template to a generic name of `application`
* regenerated questionnaire tests

PR #204, opened 2021-05-26
Add template dependency

sinclairs:
Removes the template JSON and replaces it with the [q-templates-application](https://github.com/CriminalInjuriesCompensationAuthority/q-templates-application) dependency.

Note: It's not a long term goal for templates to be handled in this way.


BarryPiccinni (approved):

sinclairs:
Also closes #143

PR #210, opened 2021-06-11
Add contextualised errors

BarryPiccinni:
Contextualise schema before validation allowing for meaningful errors to be returned.


sinclairs (approved):
Looks good for the first pass. We should probs return the section resource in the next iteration, however that would require a minor update to CW too.

PR #211, opened 2021-06-24
Move Slack Messaging service to catch block

armoj:
The Slack Service `sendMessage` function was not called when there was a communication issue between DCS and the MB. This change now catches all request errors and invokes the Slack Service's `sendMessage` accordingly.

### DIFF
--- a/questionnaire/questionnaire-service.js
+++ b/questionnaire/questionnaire-service.js
@@ -117,18 +117,20 @@ function createQuestionnaireService({
                 !submissionResponse.body ||
                 submissionResponse.body !== 'Message sent'
             ) {
-                await updateQuestionnaireSubmissionStatus(questionnaireId, 'FAILED');
-                const slackService = createSlackService();
-                slackService.sendMessage({
-                    appReference: `${process.env.APP_ENV || 'dev'}.reporter.webhook`,
-                    messageBodyId: 'message-bus-down',
-                    templateParameters: {
-                        timeStamp: new Date().getTime()
-                    }
-                });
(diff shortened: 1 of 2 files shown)


questionnaire/questionnaire-service.js
```diff
@@ -117,18 +117,20 @@ function createQuestionnaireService({
...
-                slackService.sendMessage({
-                    appReference: `${process.env.APP_ENV || 'dev'}.reporter.webhook`,
-                    messageBodyId: 'message-bus-down',
-                    templateParameters: {
-                        timeStamp: new Date().getTime()
-                    }
-                });
+                throw new VError(
+                    `Questionnaire with id "${questionnaireId}" was not successfully added to the Message Bus queue`
+                );
```
BarryPiccinni, line 122: In the interest of best practice, should this function not return something, or perhaps be structured better? It awaits a response to the post, errors if it doesn't like the result, otherwise nothing?

armoj: what do you suggest it should return?

BarryPiccinni: `submissionResponse` should probably be returned, as it fulfils the main purpose of this function.

armoj: That is the only useful that could be returned but I don't see much of a point in doing that other than to adhere to the standard of having an explicit return. This functions purpose is to kick off a submission or throw and fail. Returning anything to the caller won't help here (as it stands).

BarryPiccinni (approved):

PR #224, opened 2021-10-05
Dataset v2 openapi

sinclairs:
OpenAPI definition work to support V2 of dataset resource.

Some clunky bits in this PR to work around the JSON Schema / OpenAPI v3.0.2 schema mismatch issue. Happy to talk through these.


openapi/src/json-schemas/models/definitions/dataset/simple-open-question.json
```diff
@@ -0,0 +1,22 @@
...
+            "const": "simple"
+        },
+        "id": {
+            "$ref": "../question-id.json"
+        },
+        "label": {
+            "type": "string"
+        },
+        "closedQuestion": {
+            "enum": [false]
```
armoj, line 16: could we use `"const": false` here?

openapi/openapi.json
```diff
@@ -2308,113 +2319,628 @@
...
+                                                                                                                                    {
+                                                                                                                                        "type": "array",
+                                                                                                                                        "items": {}
+                                                                                                                                    }
+                                                                                                                                ]
+                                                                                                                            }
+                                                                                                                        }
+                                                                                                                    },
```
armoj, line 2730: amazing formatting, thanks GH!!!

openapi/src/openapi-src.json
```diff
@@ -600,7 +600,25 @@
                 "operationId": "getDataset",
                 "parameters": [
                     {
-                        "$ref": "#/components/parameters/questionnaireId"
+                        "name": "questionnaireId",
+                        "in": "path",
+                        "required": true,
+                        "description": "The id of the specific questionnaire instance. Format UUIDv4.",
```
armoj, line 606: what's the reasoning behind moving `questionnaireId` out of the `components` object? Could we not have both the `questionnaireId` and the `Accept-Version` in the `components` object for consistency?

armoj (changes requested):

openapi/src/json-schemas/models/definitions/dataset/simple-open-question.json
sinclairs: This is a clunky bit... for some reason speccy won't allow `const` to have a boolean, so it needs to be an enum. (OpenAPI v3.0.2 doesn't support the keyword `const`)

openapi/src/openapi-src.json
sinclairs: Another clunky bit... I had to being this inline to allow it to have a different `example` id. Speccy complained when I had it with as a `$ref` with a sibling `example` keyword.

As for the `Accept-Version` being inlined too... I did have this as a param definition, however the semvers are specific to the endpoint, so there wasn't much reusability to be had.

openapi/openapi.json
sinclairs: Erm, naughty bad GH 🤷‍♂️

openapi/src/json-schemas/models/definitions/dataset/simple-open-question.json
armoj: yeah thought so. just checking

openapi/src/openapi-src.json
```diff
@@ -631,18 +649,163 @@
...
+                                    },
+                                    "v2.0.0": {
+                                        "summary": "Request Header \"Accept-Version: 2.0.0\"",
+                                        "value": {
+                                            "data": [
+                                                {
+                                                    "type": "dataset",
+                                                    "id": "0",
+                                                    "attributes": {
+                                                        "values": [
```
BarryPiccinni, line 684: I'm wondering if we're not using the term "value" too often? Consumers of this object are going to have to navigate paths like: 

```dataset.values.forEach(value => {
    let answer = value.value;
})```

Not an huge issue but maybe something to consider?

sinclairs: @BarryPiccinni the "[summary](https://github.com/CriminalInjuriesCompensationAuthority/data-capture-service/pull/224/files#diff-2836d18055d85134d1f93a05d9d4a74048ef60ddac08dee1ffb30d8b5c3c97f7R677)" and "[value](https://github.com/CriminalInjuriesCompensationAuthority/data-capture-service/pull/224/files#diff-2836d18055d85134d1f93a05d9d4a74048ef60ddac08dee1ffb30d8b5c3c97f7R678)" properties are part of the OpenAPI definition, not the actual response. Is that what you were referring to?

armoj (approved):

PR #229, opened 2021-11-17
BLOCKER json-pointer vulnerability replaced with jsonpointer

neil-stephen-mcgonigle:
…ced with jsonpointer


questionnaire/questionnaire-service.js
```diff
@@ -167,7 +170,10 @@ function createQuestionnaireService({
                 }
 
                 if (task.type === 'sendSms') {
-                    if (pointer.has(questionnaire, task.templatePlaceholderMap.phoneNumber)) {
+                    if (
+                        pointer.get(questionnaire, task.templatePlaceholderMap.phoneNumber) !==
+                        undefined
```
neil-stephen-mcgonigle, line 175: not sure what's going on with the formatting?

armoj: The max line width is 80 columns. So it'll wrap this

neil-stephen-mcgonigle: Cheers, I thought it was the line wrap style I am just not used to how it looks, 
pointer.get(questionnaire, task.templatePlaceholderMap.phoneNumber)
       !== undefined
[Checkstyle](https://checkstyle.sourceforge.io/apidocs/index.html?com/puppycrawl/tools/checkstyle/checks/whitespace/OperatorWrapCheck.html)

neil-stephen-mcgonigle: as opposed to [JavaScript Style Guide](https://www.w3schools.com/js/js_conventions.asp)

package-lock.json
```diff
@@ -2486,6 +2486,14 @@
                 "json-to-ast": "^2.0.3",
                 "jsonpointer": "^4.0.1",
                 "leven": "^3.1.0"
+            },
```
armoj, line 2489: There are still references to `json-pointer` in the package-lock. Not sure if that is perfectly normal or not in this type of situation. Did you `npm uninstall` the old `json-pointer` module?

neil-stephen-mcgonigle: I did use `npm uninstall` let me have a look

neil-stephen-mcgonigle:
Not sure about the target branch for the merge `l10n-rc-temp`??

package-lock.json
neil-stephen-mcgonigle: is this due to `json-pointer` being a [dependency ](https://github.com/CriminalInjuriesCompensationAuthority/data-capture-service/blob/f4ee58564d99e7a955d1ce089a9aea1ecec46a5f/package-lock.json#L9582) of `json-translator`?

package.json
```diff
@@ -27,7 +27,7 @@
         "express-openapi-validator": "^1.6.2",
         "got": "^9.6.0",
         "helmet": "^3.21.2",
-        "json-translator": "github:CriminalInjuriesCompensationAuthority/json-translator#v0.1.0",
+        "json-translator": "github:CriminalInjuriesCompensationAuthority/json-translator",
```
neil-stephen-mcgonigle, line 30: happy to discuss it this is the correct approach?

package-lock.json
neil-stephen-mcgonigle: image
dependencies of redoc and openapi-sampler

package.json
armoj: We typically point to a specific version and don't necessarily reference the latest cut of the code.

package-lock.json
armoj: if it is due to it being a dep of `json-translator` then you would need to reinstall the translator on here because your latest version of the `json-translator` does not have `json-pointer` as a dependency.

I thing that makes sense :D

armoj: I'm not fully understanding this. Happy to call you about it. Those listed deps shouldn't be there if they are listed as deps of the now-updated json-translator module. Unless I am misunderstanding the issue in the first place!

PR #230, opened 2021-11-17
Json pointer replace

neil-stephen-mcgonigle:


neil-stephen-mcgonigle:
best way to test this?

questionnaire/questionnaire-service.js
```diff
@@ -147,7 +147,10 @@ function createQuestionnaireService({
                 const notifyService = createNotifyService({logger});
 
                 if (task.type === 'sendEmail') {
-                    if (pointer.has(questionnaire, task.templatePlaceholderMap.emailAddress)) {
+                    if (
```
sinclairs, line 150: Can we cache the value as per the replace-json-pointer "has" check?

questionnaire/questionnaire-service.js
```diff
@@ -167,7 +170,10 @@ function createQuestionnaireService({
                 }
 
                 if (task.type === 'sendSms') {
-                    if (pointer.has(questionnaire, task.templatePlaceholderMap.phoneNumber)) {
+                    if (
+                        pointer.get(questionnaire, task.templatePlaceholderMap.phoneNumber) !==
```
sinclairs, line 174: As above, can we cache the value as per the replace-json-pointer "has" check?

sinclairs (changes requested):
Looks good.

When we're happy, before merging, can you remove the "bump version to 3.0.2" commit and instead run `npm version patch`. It'll create the version commit for you and also make the corresponding change in `package-lock.json`'s version too

questionnaire/questionnaire-service.js
neil-stephen-mcgonigle: np

questionnaire/questionnaire-service.js
neil-stephen-mcgonigle: np

neil-stephen-mcgonigle:
> Looks good.
> 
> When we're happy, before merging, can you remove the "bump version to 3.0.2" commit and instead run `npm version patch`. It'll create the version commit for you and also make the corresponding change in `package-lock.json`'s version too

cool

package-lock.json
```diff
@@ -2486,6 +2486,14 @@
                 "json-to-ast": "^2.0.3",
                 "jsonpointer": "^4.0.1",
                 "leven": "^3.1.0"
+            },
```
sinclairs, line 2489: File is missing version bump

sinclairs:
> 
> 
> best way to test this?

I'd hope the current tests cover this, you reckon we need something else?

neil-stephen-mcgonigle:
> > best way to test this?
> 
> I'd hope the current tests cover this, you reckon we need something else?
Happy with that just wanted assurance.

sinclairs (changes requested):
Nice one. Just need to sort the commit order and we're good.

Current order: 
- Replace json-pointer module with jsonpointer
- 3.0.3
- caching calls to pointer.get 

Can we squash? 
- Replace json-pointer module with jsonpointer
- caching calls to pointer.get 

Proposed order (oldest to newest):
- Replace json-pointer module with jsonpointer
- 3.0.3 

We can then tag the 3.0.3 as the release

sinclairs (approved):

PR #231, opened 2021-11-18
replace json-pointer with jsonpointer

neil-stephen-mcgonigle:


neil-stephen-mcgonigle:
Apologies, I went into the manual merge stage on GitHub and it's created a new commit which we don't want. I'll fix it after the drilling stops.

PR #236, opened 2021-11-25
Notify integration test

sinclairs:
Adds an integration test for the current notify service. This is in preparation for a major version bump to `notifications-node-client`.

Adds a `test:local` npm script which allows the integration test to be excluded.

Update: Now has the notifications-node-client 5.1.0

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -32,7 +32,7 @@ config.coverageThreshold = {
         lines: 9,
         statements: 9
     },
-    './*/!(message-bus|notify|request)/**/*.js': {
+    './*/!(message-bus|request)/**/*.js': {
         statements: 54,
         branches: 60,
         functions: 50,
@@ -43,12 +43,6 @@ config.coverageThreshold = {
         branches: 60,
         functions: 0,
         lines: 15
-    },
-    './*/notify/**/*.js': {
-        statements: 42,
-        branches: 60,
-        functions: 0,
-        lines: 42
     },
     './*/request/**/*.js': {
         statements: 23,
--- a/services/notify/index.js
+++ b/services/notify/index.js
@@ -5,24 +5,55 @@ const createMessageBusCaller = require('../message-bus');
 
 const {NotifyClient} = notificationNodeClient;
 const apiKey = process.env.NOTIFY_API_KEY;
-const notifyClient = new NotifyClient(apiKey);
+const sharedNotifyClient = new NotifyClient(apiKey);
 
 function createNotifyService(spec) {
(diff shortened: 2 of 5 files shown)


jest.config.js
```diff
@@ -7,12 +7,6 @@ const config = {
 };
 /*! m0-end */
 
-process.env.DCS_JWT_SECRET = '123';
-process.env.DCS_LOG_LEVEL = 'silent';
-process.env.NOTIFY_API_KEY = '123apiKey';
-process.env.WEBHOOK_DEV_REPORTER_SLACK = 'webhook-url';
-process.env.APP_ENV = 'test';
```
armoj, line 14: Are none of these being used at all in tests?

services/notify/index.js
```diff
@@ -10,19 +10,29 @@ const notifyClient = new NotifyClient(apiKey);
...
+            return {
+                id: smsSendRequest.body.id
+            };
         } catch (err) {
-            logger.error({err}, 'SMS SEND FAILURE');
+            logger.error(err.error, 'SMS SEND FAILURE');
         }
 
-        return response;
+        return undefined;
```
armoj, line 35: would it be worth having the return values in the same shape here, do you think?

armoj (changes requested):

.gitignore
```diff
@@ -69,3 +69,5 @@ typings/
 .vscode
 
 .config
+
+jest.env.js
```
armoj, line 73: just spotted this. Ignore my comment about the env vars

armoj (changes requested):

jest.config.js
sinclairs: Closed as per: https://github.com/CriminalInjuriesCompensationAuthority/data-capture-service/pull/236#discussion_r757049178

services/notify/index.js
sinclairs: Possibly. I was just maintaining the current behaviour. Nothing uses it at the moment.

services/notify/index.js
```diff
@@ -5,24 +5,55 @@ const createMessageBusCaller = require('../message-bus');
...
     async function sendSms(options) {
-        let response;
         try {
-            response = notifyClient.sendSms(options.templateId, options.phoneNumber, {
-                personalisation: {
-                    case_reference: options.personalisation.caseReference
-                },
-                reference: null
-            });
+            const smsSendRequest = await notifyClient.sendSms(
```
armoj, line 33: `notificationDocument`
`notificationReceipt`

PR #239, opened 2021-12-06
Add get attributes from section functionailty

sinclairs:
Both the dataset endpoint and the answer summary page require getting attributes from sections. This PR provides that functionality.

All tests passing.

### DIFF
--- a/questionnaire/section/section.test.js
+++ b/questionnaire/section/section.test.js
@@ -1,0 +1,220 @@
+'use strict';
+
+const fixtures = require('./test-fixtures');
+const createSection = require('./section');
+
+describe('Section', () => {
+    describe('getAttributesByData', () => {
+        describe('Given a section definition containing a schema', () => {
+            describe('And the schema contains an attribute that can accept a single value', () => {
+                it('should return a simple attribute with a label and single value', async () => {
+                    const section = createSection({
+                        sectionDefinition: fixtures.sectionWithSimpleAttributeSingleValue
+                    });
+                    const data = {
+                        'q-applicant-enter-your-email-address':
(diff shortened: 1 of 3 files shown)


questionnaire/section/section.test.js
```diff
@@ -0,0 +1,220 @@
...
+const fixtures = require('./test-fixtures');
+const createSection = require('./section');
+
+describe('Section', () => {
+    describe('getAttributesByData', () => {
+        describe('Given a section definition containing a schema', () => {
+            describe('And the schema contains an attribute that can accept a single value', () => {
+                it('should return a simple attribute with a label and single value', async () => {
+                    const section = createSection({
+                        sectionDefinition: fixtures.sectionWithSimpleAttributeSingleValue
```
armoj, line 12: just noticed this function signature too. Objects all the way!

questionnaire/section/section.test.js
```diff
@@ -0,0 +1,220 @@
...
+                    it('should return an empty list of attributes', async () => {
+                        const section = createSection({
+                            sectionDefinition: fixtures.sectionWithSimpleAttributeSingleValue
+                        });
+                        const data = {
+                            'this-key-does-not-exist-in-the-section-schema': 'dummy value'
+                        };
+                        const sectionAttributes = section.getAttributesByData(data);
+
+                        expect(sectionAttributes).toEqual([]);
```
armoj, line 92: Would it be worth having a test for a data object that contains at least one genuine answer and 1 dummy one?

sinclairs: I reckon it's ok. There is test for ignoring non-existent keys and another for returning keys that exist, and a another that returns 2 out of 3 things. Pretty much everything is covered in the coverage report. I can add one if you like, but I don't think it's going to increase coverage

PR #240, opened 2021-12-09
Implement dataset v2

sinclairs:
Updates the dataset service to allow version 2.0.0 of the dataset resource to be returned.

Tests have been refactored for the new functionality.

### DIFF
--- a/questionnaire/dataset/dataset-service.js
+++ b/questionnaire/dataset/dataset-service.js
@@ -4,15 +4,66 @@ const VError = require('verror');
 
 const defaults = {};
 defaults.createQuestionnaireDAL = require('../questionnaire-dal');
+defaults.createSection = require('../section/section');
 
 function createDatasetService({
     logger,
-    createQuestionnaireDAL = defaults.createQuestionnaireDAL
+    createQuestionnaireDAL = defaults.createQuestionnaireDAL,
+    createSection = defaults.createSection
 } = {}) {
     const db = createQuestionnaireDAL({logger});
 
-    async function getResource(questionnaireId) {
-        const questionnaire = await db.getQuestionnaire(questionnaireId);
+    function mergeArrays(target, source) {
+        const targetIsArray = Array.isArray(target);
+        const sourceIsArray = Array.isArray(source);
+
+        if (targetIsArray && sourceIsArray) {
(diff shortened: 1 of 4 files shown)


questionnaire/dataset/dataset-service.js
```diff
@@ -21,42 +72,84 @@ function createDatasetService({
...
+            if (questionAnswers !== undefined) {
+                const sectionDefinition = sections[sectionId];
+                const section = createSection({sectionDefinition});
+                const attributes = section.getAttributesByData(questionAnswers);
+
+                attributes.forEach(attribute => {
+                    const existingAttribute = dataset.get(attribute.id);
+
+                    if (existingAttribute !== undefined) {
+                        mergeAttributeValues(existingAttribute, attribute);
```
armoj, line 107: is this the functionality specifically for the merging physical injuries answers from multiple pages?

questionnaire/dataset/dataset-service.js
```diff
@@ -21,42 +72,84 @@ function createDatasetService({
...
+                    } else {
+                        const mutatedAttribute = applyGlobalAttributeDetails(
```
armoj, line 109: are the code blocks within this if/else always mutually exclusive? I must not be understanding this logic as it appears to be the `if` deals with the munging of the phy-inj stuff... and the `else` deals with the retrieval of the global meta title (also related to phy-inj.

questionnaire/dataset/dataset-service.js
sinclairs: Yeah, just any attributes that might be captured across multiple pages. Only physical injuries at the moment

questionnaire/dataset/dataset-service.js
sinclairs: @armoj If there is an existing attribute it'll merge them, else it'll apply any global attribute data that has been defined.

So for physical injuries, it'll get the global label initially. Then when it encounters it again it'll merge the values.

armoj: Ok makes sense. So this code programmatically assumes that if there is a "global meta title", then there will be more than 1 attribute to merge together here (e.g. exactly like the physical injuries

PR #268, opened 2022-02-25
Update JWT secret key to allow multiple secrets

BarryPiccinni:
Allow multiple JWTs to be used


neil-stephen-mcgonigle (approved):

armoj:
is this actually needing to go to master? I thought we were just going to deploy this branch, and then redeploy master when the new secrets are rolled out

PR #298, opened 2022-06-23
Personalisation

sinclairs:


questionnaire/questionnaire/questionnaire.test.js
```diff
@@ -344,6 +526,8 @@ const questionnaireDefinition = {
         'p-applicant-british-citizen-or-eu-national',
         'p-applicant-theme-undefined',
         'p-applicant-theme-not-found',
+        // 'p--has-legal-authority',
+        // 'p-applicant-are-you-18-or-over',
```
armoj, line 530: I this to explicitly explain that a test should fail due the absence of these array items?

sinclairs: No! Some erroneous comments. I'll remove those 👍

armoj (approved):
happy with this

PR #299, opened 2022-06-24
Fix audit for got module

sinclairs:


services/request/index.js
```diff
@@ -11,24 +11,27 @@ function createRequestService() {
                 accept: 'application/vnd.api+json',
                 'Content-Type': 'application/vnd.api+json'
             },
-            json: true,
-            body: options.body
```
BarryPiccinni, line 15: This is what I was trying (poorly) to describe on Friday.  By removing the explictly set `body: options.body`, but still merging this object with `options`, means you will still have something like
```
{
  options: {
    body: <payload>
  }
} 
```

Rather than the suggested

```
{
  options: {
    json: <payload>
  }
}
```

jest.config.js
```diff
@@ -40,16 +40,10 @@ config.coverageThreshold = {
...
-    },
-    './*/request/**/*.js': {
-        statements: 23,
-        branches: 60,
-        functions: 0,
-        lines: 23
+        statements: 100,
+        branches: 100,
+        functions: 100,
+        lines: 100
```
BarryPiccinni, line 46: The coverage for the message bus and request service are now well above the default, they don't need explicit limits anymore. Consider removing lines 36-48 instead.

BarryPiccinni (approved):

BarryPiccinni (approved):

PR #318, opened 2023-01-16
Add metadata endpoint

BarryPiccinni:


sinclairs (changes requested):

openapi/src/json-schemas/api/resources/metadata-resource-v1.0.0.json
```diff
@@ -0,0 +1,35 @@
...
+    "title": "Metadata resource v1.0.0",
+    "allOf": [
+        {"$ref": "../json-api-base-resource.json"},
+        {
+            "properties": {
+                "type": {
+                    "const": "metadata"
+                },
+                "id": {
+                    "const": "0"
```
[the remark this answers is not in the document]
BarryPiccinni: I have changed this to be a questionnaire Id as suggested.

openapi/openapi.json
```diff
@@ -3362,6 +3362,229 @@
...
+                                                                                "title": "Any valid resource"
+                                                                            },
+                                                                            "relationships": {
+                                                                                "type": "object"
+                                                                            }
+                                                                        }
+                                                                    },
+                                                                    {
+                                                                        "properties": {
+                                                                            "type": {
```
armoj, line 4040: is it worth employing `"additionalProperties": true`, or `patternProperties` here or anything. Make it is strict/loose as needed and still allow arbitrary key/values

armoj (changes requested):
What is the logic behind the version 1.0.0/2.0.0 of the resource for meta? I remember it being used for the dataset endpoint due to there needing to be 2 version of the response for that endpoint, but not sure what its purpose is here. This is just a question, not a request to alter it

openapi/src/openapi-src.json
```diff
@@ -228,6 +228,101 @@
...
+                "description": "Gets the answer(s) to a supplied section",
+                "operationId": "getAnswer",
+                "parameters": [
+                    {
+                        "$ref": "#/components/parameters/questionnaireId"
+                    },
+                    {
+                        "$ref": "#/components/parameters/sectionId"
+                    },
+                    {
```
sinclairs, line 244: Given the all answers endpoint isn't versioned, it might be worth leaving off the versioning for the first pass

sinclairs (changes requested):

sinclairs:
> What is the logic behind the version 1.0.0/2.0.0 of the resource for meta? I remember it being used for the dataset endpoint due to there needing to be 2 version of the response for that endpoint, but not sure what its purpose is here. This is just a question, not a request to alter it

https://cloud.google.com/blog/products/api-management/api-design-which-version-of-versioning-is-right-for-you

It allows us to change the resource over time. Although I think the guidance above does suggest that it might not always be useful.

openapi/src/json-schemas/api/resources/metadata-resource-v1.0.0.json
```diff
@@ -0,0 +1,35 @@
...
+                },
+                "attributes": {
```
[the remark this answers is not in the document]
armoj: In order to avoid having to use quotes for object property names?

BarryPiccinni:
@sinclairs I've fixed the examples in my latest push. I've also added a second filter component as it was specific to progress entries and so didn't make sense when including it in the meta endpoint

openapi/src/json-schemas/models/definitions/timestamp.json
```diff
@@ -0,0 +1,6 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "Timestamp",
+    "type": "string",
+    "pattern": "^(?:[1-9]\\d{3}-(?:(?:0[1-9]|1[0-2])-(?:0[1-9]|1\\d|2[0-8])|(?:0[13-9]|1[0-2])-(?:29|30)|(?:0[13578]|1[02])-31)|(?:[1-9]\\d(?:0[48]|[2468][048]|[13579][26])|(?:[2468][048]|[13579][26])00)-02-29)T(?:[01]\\d|2[0-3]):[0-5]\\d:[0-5]\\d(?:\\.\\d{1,9})?(?:Z|[+-][01]\\d:[0-5]\\d)$"
```
sinclairs, line 5: JSON Schema already has this baked in e.g. https://github.com/CriminalInjuriesCompensationAuthority/q-templates-application/blob/b40fa7ce73d52c3051acd1b1f1db8fffafa30a61/src/lib/resource/sections/applicant-enter-your-date-of-birth.js#L59

Do we need this regex?

sinclairs (changes requested):

openapi/src/json-schemas/models/definitions/timestamp.json
BarryPiccinni: I was quite pleased with how complex it was, but no we don't need it! Have changed it to the JSON schema format.

armoj:
Adding here for visibility:

For the metadata DCS implementation, in the interest of separation of concerns... I am thinking it would be beneficial to separate out the metadata functionality in to its own directory (similarly to the dataset stuff).

`/questionnaire/metadata`
`/questionnaire/metadata/route.js` Have your endpoints defined in here and link to it [like this](https://github.com/CriminalInjuriesCompensationAuthority/data-capture-service/blob/ddae84a802ec8a85543c8b2f2c4cc2c59497de12/questionnaire/routes.js#L41) in the questionnaire routes.js
`/questionnaire/metadata/metadata-service.js` (have the logic to get/build the resource(s) in here. You can also then have a util/helper function that will add 31 days to the created date with no need to "one line" it. Better for testability also.

BarryPiccinni:
> You can also then have a util/helper function that will add 31 days to the created date with no need to "one line" it. Better for testability also.

I didn't do this as the `expires` function is only a first pass until we store it in the DB and reference it as we do the rest of the metadata.

questionnaire/questionnaire-dal.js
```diff
@@ -177,6 +177,34 @@ function questionnaireDAL(spec) {
...
+            } else {
+                result = await db.query(
+                    "SELECT id, questionnaire -> 'meta' -> 'questionnaireDocumentVersion' AS \"questionnaire-version\", created, modified, submission_status, questionnaire -> 'answers' -> 'user' -> 'user-id' AS \"user-id\" FROM questionnaire"
+                );
+            }
+        } catch (err) {
+            throw err;
+        }
+
+        return result;
```
armoj, line 205: .map() this to remove all other redundant data in the response. Just return an array of objects

armoj (changes requested):

questionnaire/routes.js
```diff
@@ -105,6 +107,18 @@ router
...
         }
+    })
+    .get(permissions('read:questionnaires'), async (req, res, next) => {
+        try {
+            const questionnaireService = createQuestionnaireService({logger: req.log});
+            const response = await questionnaireService.getAnswersBySectionId(
+                req.params.questionnaireId,
+                req.params.sectionId
+            );
+            res.status(200).json(response);
```
armoj, line 118: change to:
````js
res.status(200).json({
    data: response
});
````

**Review by armoj — CHANGES_REQUESTED** [review]

File: package.json [hunk 8]
```diff
@@ -1,6 +1,6 @@
 {
     "name": "data-capture-service",
-    "version": "6.2.6",
```
sinclairs, line 3: As this is adding new functionality in a backwards compatible way, this should be a minor release rather than a patch.

If it was behind a toggle the patch would apply

questionnaire/questionnaire-dal.js
```diff
@@ -177,6 +177,34 @@ function questionnaireDAL(spec) {
...
+                    throw new VError(
+                        {
+                            name: 'ResourceNotFound'
+                        },
+                        `User "${query.filter.userId}" not found`
+                    );
+                }
+            } else {
+                result = await db.query(
+                    "SELECT id, questionnaire -> 'meta' -> 'questionnaireDocumentVersion' AS \"questionnaire-version\", created, modified, submission_status, questionnaire -> 'answers' -> 'user' -> 'user-id' AS \"user-id\" FROM questionnaire"
```
armoj, line 198: should this else block use the same pattern as the preceding if block above it?: get result, throw if rowCount is 0. Then return at the bottom of the function if it falls through to it

armoj (changes requested):

questionnaire/routes.test.js
```diff
@@ -188,3 +188,44 @@ describe('Issue: https://github.com/cdimascio/express-openapi-validator/issues/7
...
+                    res.body.included[0].attributes.properties.confirmation.description;
+
+                expect(confirmation).toMatch(/19\\751194/);
+                expect(confirmation).toMatch(/Barry\.foo@foo\.com/);
+                expect(res.statusCode).toBe(200);
+                expect(res.type).toBe('application/vnd.api+json');
+            });
+        });
+    });
+});
```
sinclairs, line 231: Can this test live with the other metadata routing tests? Also, I'm not quite following what the metadata bit of this test is?

questionnaire/metadata/metadata-service.test.js
```diff
@@ -0,0 +1,49 @@
...
+                        type: 'metadata',
+                        id: '6fbc0958-e70c-4e37-ab65-12b96e8d37c1',
+                        attributes: {
+                            'questionnaire-id': '6fbc0958-e70c-4e37-ab65-12b96e8d37c1',
+                            'questionnaire-document-version': '4.2.0',
+                            created: '2023-01-01T10:42:24.016Z',
+                            modified: '2023-01-01T10:42:49.604Z',
+                            state: 'NOT_STARTED',
+                            'user-id': 'abc123',
+                            expires: '2023-02-01T00:00:00.000Z'
```
sinclairs, line 40: The keys in the response seem to differ from those returned from the DAL e.g. `submission_status` to `state`. I thought they'd be the same? `submission-status` is more descriptive

questionnaire/metadata/metadata-routes.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+const express = require('express');
+
+const permissions = require('../../middleware/route-permissions');
+const createMetadataService = require('./metadata-service');
+
+const router = express.Router();
+
+router.route('/meta').get(permissions('read:questionnaires'), async (req, res, next) => {
```
sinclairs, line 10: For consistency, should all "meta" refs be replaced with "metadata"

sinclairs (changes requested):
Looking good, just a few small things

questionnaire/questionnaire-dal.js
BarryPiccinni: Well the thinking was to get to this `else` branch the user has asked for all metadata. The result is either some metadata or no metadata. They've not provided any parameters that would generate an error.

questionnaire/questionnaire-dal.js
```diff
@@ -177,6 +177,34 @@ function questionnaireDAL(spec) {
...
+                result = await db.query(
+                    `SELECT id, questionnaire -> 'meta' -> 'questionnaireDocumentVersion' AS "questionnaire-version", created, modified, submission_status, questionnaire -> 'answers' -> 'user' -> 'user-id' AS "user-id" FROM questionnaire WHERE questionnaire -> 'answers' -> 'user' ->> 'user-id' = $1`,
+                    [query.filter.userId]
+                );
+                if (result.rowCount === 0) {
+                    throw new VError(
+                        {
+                            name: 'ResourceNotFound'
+                        },
+                        `User "${query.filter.userId}" not found`
```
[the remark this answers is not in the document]
BarryPiccinni: I feel like the error is derived from the parameter here. Metadata can exist or there is no metadata - both are valid results - but a dodgy userId should throw an error. We don't currently distinguish between a dodge userId and 0 rows tho, so I can see both sides of this.

package.json
BarryPiccinni: This is being merged into an RC, so I've removed all versioning for now. We can version the RC at deployment.

armoj (approved):

PR #326, opened 2023-03-27
Bp govuk auth header

BarryPiccinni:


questionnaire/dataset/dataset-routes.js
```diff
@@ -17,6 +17,7 @@ router
 
             const resourceCollection = await datasetService.getResource(
                 questionnaireId,
+                req.header('On-behalf-of'),
```
armoj, line 20: the header name is sentence case here, whereas it is all lowercase in the spec. Can we make that consistent, which ever is chosen

questionnaire/dataset/dataset-service.js
armoj, line 1: userId/user-id is used throughout these files. Are we wanting to use ownerId/owner-id instead as discussed?

armoj (changes requested):

PR #329, opened 2023-04-06
CDBT - 24

msenkiw:


services/notify/message-bus/index.external.integration.test.js
```diff
@@ -2,7 +2,7 @@
 
 const createNotifyService = require('./index');
 
-describe('Notify service', () => {
+describe.skip('Notify service', () => {
```
jack-burt-is, line 5: Is there a reason we're skipping tests rather than deleting them if they've become unnecessary due to removing functionality?

jack-burt-is (commented):
Some initial thoughts from me to consider before getting CICA's view

BarryPiccinni (commented):
I will go thru the changes more thoroughly, but we'll need to do some house keeping on this branch before we can deploy it live, so in the interest of not being a blocker, here's some initial thoughts:

- The Slack integration needs to be fixed. It was commented out because it was throwing an error which was preventing the branch being deployed. I assume it's because something being returned from the SQS module didn't match what was being returned by the old module but I didn't investigate that.
- The comments and commit messages will need to be tidied up. Some of them were my WIPs when I was just trying to get the thing working - I wouldn't have expected these to necessarily make it into the code base. Should do a quick review of the changes and fixup ones that are unnecessary.
- The tests are failing which will prevent the branch from deploying.

PR #330, opened 2023-04-06
Update createQuestionnaire contract to include owner data

BarryPiccinni:
***NEW RC***


sinclairs (changes requested):
Looking good. Do we need the on-behalf-of header when creating questionnaires, given it's now being posted in the body. Could simplify things?

openapi/src/json-schemas/api/_questionnaires/post/req/201.json
```diff
@@ -15,9 +15,23 @@
                 "attributes": {
                     "type": "object",
                     "additionalProperties": false,
+                    "required": ["templateName", "owner"],
                     "properties": {
                         "templateName": {
                             "$ref": "../../../../models/definitions/questionnaire-template-name.json"
+                        },
+                        "owner": {
```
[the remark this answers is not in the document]
BarryPiccinni: I've pulled the id into it's own file for reuse, off the top of my head I don't think it'll be necessary to use the 'owner' object anywhere else.

openapi/openapi.json
```diff
@@ -73,7 +77,11 @@
                                     "example": {
                                         "data": {
                                             "attributes": {
-                                                "templateName": "this-does-not-exist"
+                                                "templateName": "this-does-not-exist",
+                                                "owner": {
+                                                    "id": "urn:uuid:f81d4fae-7dec-11d0-a765-00a0c91e6bf6",
```
armoj, line 82: https://www.rfc-editor.org/rfc/rfc2141#section-2

instead of `uuid` as the namespace (it is more of a format in my eyes), could we have an identifier for our service/issuer. Something like:

`urn:cica:56P4CMsGh_02YOlWpd8PAOI-2sVlB2nsNU7mcLZYhYw=`

perhaps we should follow [One Logins example](https://docs.sign-in.service.gov.uk/integrate-with-integration-environment/process-identity-information/#process-your-user-s-identity-information) and use `fdc` described [here](https://www.iana.org/assignments/urn-namespaces/urn-namespaces.xhtml)

openapi/openapi.json
```diff
@@ -103,11 +111,28 @@
...
+                                                    "owner": {
+                                                        "type": "object",
+                                                        "additionalProperties": false,
+                                                        "required": ["id", "isAuthenticated"],
+                                                        "properties": {
+                                                            "id": {
+                                                                "title": "Owner Id",
+                                                                "type": "string",
+                                                                "format": "uri",
+                                                                "pattern": "^urn:(uuid|fdc:gov.uk:2022):"
```
armoj, line 130: have a max length on this regex to stop huge strings. Whole string length should have a max of 64, I think!

questionnaire/routes.test.js
```diff
@@ -188,3 +188,149 @@ describe('Issue: https://github.com/cdimascio/express-openapi-validator/issues/7
...
+            expect(response.body).toHaveProperty('errors');
+            expect(response.body.errors[0].status).toEqual(401);
+            expect(response.body.errors[0].detail).toEqual('jwt malformed');
+        });
+
+        it('should return status code 403 if the bearer token has insufficient scope', async () => {
+            // THIS IS A TOKEN WITH A DUMMY SCOPE
+            const dummyToken =
+                'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJhdWQiOiJkYXRhLWNhcHR1cmUtc2VydmljZSIsImlzcyI6IiQuYXVkIiwianRpIjoiNTFhODljYWUtM2Q1MC00ZDc1LTliMmEtMjU2NzliODgwMTkxIiwic3ViIjoiY2ljYS13ZWIiLCJzY29wZSI6ImNyZWF0ZTpub3RoaW5nIiwiaWF0IjoxNjgwNzk4NDU5fQ.97LgtlW_dcAV0Xno6BsbVmuyhLtq4gCoVWGQ56_VmEk';
+            // eslint-disable-next-line global-require
```
armoj, line 245: instead of these comments all throughout the file, disable for whole file by using `/* eslint-disable global-require */` at the very top of the file.

armoj (changes requested):

openapi/src/openapi-src.json
```diff
@@ -945,6 +957,18 @@
                 "example": {
                     "before": "p-applicant-declaration"
                 }
+            },
+            "onBehalfOf": {
```
sinclairs, line 961: I think this can be removed for this PR now?

sinclairs (changes requested):
@BarryPiccinni 2 small things and we're ready to go

PR #331, opened 2023-04-13
Add owner to createQuestionaire function and db

BarryPiccinni:


questionnaire/routes.test.js
```diff
@@ -266,40 +272,37 @@ describe('POST /questionnaires', () => {
...
 
         it('should return status code 201 if owner data is included in the request body', async () => {
-            jest.doMock('./questionnaire-service.js', () =>
-                jest.fn(() => ({
-                    createQuestionnaire: () => {
-                        return {
-                            type: 'questionnaires',
-                            id: '285cb104-0c15-4a9c-9840-cb1007f098fb',
-                            attributes: {
+            jest.isolateModules(() => {
```
armoj, line 285: This isn't necessarily an incorrect usage. It indicates to me that set up and tear down of mocks is not working as expected elsewhere.

armoj (changes requested):

PR #343, opened 2023-05-18
Bp add create questionnaire code to v2

BarryPiccinni:
Add owner data to create questionnaire code for openapi V2.

### DIFF
--- a/questionnaire/questionnaire-service.js
+++ b/questionnaire/questionnaire-service.js
@@ -20,6 +20,7 @@ defaults.createQuestionnaireDAL = require('./questionnaire-dal');
 
 function createQuestionnaireService({
     logger,
+    apiVersion,
     createQuestionnaireDAL = defaults.createQuestionnaireDAL
 } = {}) {
     const db = createQuestionnaireDAL({logger});
@@ -34,7 +35,7 @@ function createQuestionnaireService({
 
     ajv.addFormat('mobile-uk', ajvFormatsMobileUk);
 
-    async function createQuestionnaire(templateName) {
+    async function createQuestionnaire(templateName, ownerData) {
         if (!(templateName in templates)) {
             throw new VError(
                 {
@@ -46,6 +47,24 @@ function createQuestionnaireService({
 
         const uuidV4 = uuidv4();
         const questionnaire = templates[templateName](uuidV4);
+
+        if (apiVersion === '2023-05-17') {
+            if (!ownerData) {
+                throw new VError(
+                    {
+                        name: 'OwnerNotFound'
+                    },
+                    `Owner data must be defined`
+                );
+            }
+
+            questionnaire.answers = {
+                owner: {
+                    ownerId: ownerData.id,
+                    isAuthenticated: ownerData.isAuthenticated
+                }
+            };
+        }
 
         await db.createQuestionnaire(uuidV4, questionnaire);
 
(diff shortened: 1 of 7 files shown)


questionnaire/questionnaire-service.js
```diff
@@ -47,6 +49,25 @@ function createQuestionnaireService({
...
+                        name: 'OwnerNotFound'
+                    },
+                    `Owner data must be defined`
+                );
+            }
+
+            questionnaire.answers = {
+                ...questionnaire.answers,
+                owner: {
+                    'owner-id': owner.id,
```
sinclairs, line 65: I know we had a chat around the hyphen stuff... is it worth reverting to `ownerId` when storing it on the questionnaire?

questionnaire/questionnaire-service.js
```diff
@@ -47,6 +49,25 @@ function createQuestionnaireService({
...
+                throw new VError(
+                    {
+                        name: 'OwnerNotFound'
+                    },
+                    `Owner data must be defined`
+                );
+            }
+
+            questionnaire.answers = {
+                ...questionnaire.answers,
```
sinclairs, line 63: This is the first ever answers in the questionnaire lifecycle, so we can drop the spread

armoj (changes requested):

questionnaire/questionnaire-service.js
```diff
@@ -20,6 +20,8 @@ defaults.createQuestionnaireDAL = require('./questionnaire-dal');
 
 function createQuestionnaireService({
     logger,
+    ownerId, // eslint-disable-line no-unused-vars
```
armoj, line 23: remove this. Add it in when it is relevant

questionnaire/questionnaire-service.test.js
```diff
@@ -1,7 +1,27 @@
+/* eslint-disable no-shadow */
```
armoj, line 1: Are there Supertest routes test for the Express routing that implements these functions?

questionnaire/questionnaire-service.test.js
```diff
@@ -192,26 +212,25 @@ function getQuestionnaireDefinition() {
...
-            );
-        }
-    })
-});
-
 describe('Questionnaire Service', () => {
     describe('Answering a section', () => {
+        const questionnaireService2 = createQuestionnaireService({
+            logger: () => 'Logged from dataset test',
+            createQuestionnaireDAL: () => ({
```
armoj, line 219: Can this be Mocked and just naturally passed in via the dependency injection? Not explicitly passed in to the `createQuestionnaireService` function

armoj (changes requested):

questionnaire/questionnaire-service.test.js
BarryPiccinni: I didn't change this, only moved it. I'd rather avoid refactoring old tests where I can as these will be deprecated with v1.

questionnaire/routes.js
```diff
@@ -27,10 +27,13 @@ router.route('/').post(permissions('create:questionnaires'), async (req, res, ne
...
         }
 
-        const {templateName} = req.body.data.attributes;
+        const {templateName, owner} = req.body.data.attributes;
 
-        const questionnaireService = createQuestionnaireService({logger: req.log});
-        const response = await questionnaireService.createQuestionnaire(templateName);
+        const questionnaireService = createQuestionnaireService({
+            logger: req.log,
+            apiVersion: req.get['Dcs-Api-Version']
```
armoj, line 34: Add a routing-level test to validate the header `Dcs-Api-Version` property

armoj (changes requested):

PR #344, opened 2023-05-24
Bug: fix session timeout handling post submission

neil-stephen-mcgonigle:
- refactored keep-alive tests out to own file
- added post submission sessionResource handling


questionnaire/questionnaire-service.js
```diff
@@ -350,7 +350,23 @@ function createQuestionnaireService({
         return progressEntryResource;
     }
 
+    function applicationSubmitted(submissionStatus) {
+        return submissionStatus !== 'NOT_STARTED';
```
armoj, line 354: does this _categorically_ mean that it is submitted?

questionnaire/questionnaire-service.js
```diff
@@ -350,7 +350,23 @@ function createQuestionnaireService({
...
     async function buildSessionInformationBlock(questionnaireId) {
+        const submissionStatus = await getQuestionnaireSubmissionStatus(questionnaireId);
+
+        if (applicationSubmitted(submissionStatus)) {
+            const dateNow = Date.now();
+            return {
+                alive: false,
+                duration: 0,
+                created: dateNow,
+                expires: dateNow
```
armoj, line 366: not understanding the reason behind these 3 properties. duration should not be 0, and it was created before now.

armoj (changes requested):

questionnaire/questionnaire-service.js
neil-stephen-mcgonigle: As far as I am aware the application has been submitted in the sense that the Agree and submit" button has been clicked the application now no longer has a submission_status of 'NOT_STARTED' so submission HAS started the possible states post submission application could be in are [FAILED, IN_PROGRESS, COMPLETE].  
All of which represent post submission status therefore a confirmation page will be shown and the application session will be over.

questionnaire/keep-alive.test.js
```diff
@@ -0,0 +1,340 @@
+'use strict';
+
+/* * ****************************************************************************************** * */
+/* * ****************************************************************************************** * */
+/* *         THIS FILE IS GENERATED. ALL MANUAL EDITS MADE TO THIS FILE WILL BE LOST!!!         * */
+/* *         --------------------------------------------------------------------------         * */
+/* *         If you need to make a change to this test file you will need to edit the           * */
+/* *         generate-tests file and regenerate the tests using command line.                   * */
+/* * ****************************************************************************************** * */
+/* * ****************************************************************************************** * */
```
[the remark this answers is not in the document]
neil-stephen-mcgonigle: A hangover from the copy and paste, I shall also remove from here 
https://github.com/CriminalInjuriesCompensationAuthority/data-capture-service/blob/c818f6f9e40fe132da42dcfe50a2f61930b4df95/questionnaire/questionnaire.test.js#L4

questionnaire/questionnaire-service.js
neil-stephen-mcgonigle: These properties represent the state of _this_ session resource, the duration is 0 seconds as the session has ended, the resource is being created and expired at the same time. In essence all parameters represent an invalid session expiry resource.

neil-stephen-mcgonigle: happy to discuss also happy to change this from dateNow to creationDate or equivalent

questionnaire/questionnaire-service.js
neil-stephen-mcgonigle: Made the change anyway, removed the function and added the conditional submissionStatus check within the code.

armoj (approved):

PR #360, opened 2023-06-09
Add metadata endpoints to openapi v2023-05-17

BarryPiccinni:


openapi/src/json-schemas/models/definitions/metadata-id.json
```diff
@@ -0,0 +1,6 @@
+{
```
sinclairs, line 1: What is this for?

PR #364, opened 2023-06-30
 [CDBT-138] Action to transform and upload Answers

jack-burt-is:
This covers the action to transform the raw questionnaire into the format required by the downstream components, which is roughly:
```
{
    meta: {
        caseReference: '23\\123456'
        dateSubmitted: '2023-05-19T13:06:12.693Z'
    },
    themes: [Themes],
    declaration: {
        type: 'simple',
        id: 'p-applicant-declaration',
        label: <div id="declaration">...,
        value: 'i-agree',
        valueLabel: 'I have read and understood the declaration'
    }
}
```
Where [Themes] is the check your answers output, but with amalgamated injuries which also gets handled as part of the flattenAnswers method.
___

Keen to get visibility over this from @sinclairs and @BarryPiccinni on approach. There's a few things in particular on my mind:
- I have added a getId() function onto the questionnaire object as I couldn't see another way to get that value from the object, am I missing something or is this valid?
- I wasn't sure how to inject to he logger into the new `integration` service, which is required for general logging but also for injecting into the S3 client and Questionnaire DAL. Usually it's pulled from the HTTP request I think? But not sure how that's going to integrate with the task-runner
- I borrowed some code from the dataset service (I didn't extract this out into another helper to use in both under the assumption that the dataset service is going to be retired soon anyway from my understanding so that code won't be duplicated once that happens
- Would be good just to get your view on the overall approach to the transformation, it feels overall clean and I was keen not to over-complicate it given the timescales.


sinclairs:
>     * I have added a getId() function onto the questionnaire object as I couldn't see another way to get that value from the object, am I missing something or is this valid?

`getId` LGTM 👍

>     * I wasn't sure how to inject to he logger into the new `integration` service, which is required for general logging but also for injecting into the S3 client and Questionnaire DAL. Usually it's pulled from the HTTP request I think? But not sure how that's going to integrate with the task-runner

Some route handler will kick this process off and have access to the logger that's scoped to the request. We can either make the integration service a factory and pass it in on creation, or pass it in to the runner context, something like:

```
async function transformAndUpload({questionnaire, logger}) {
    ...
    return submissionResponse;
}

...

const logger = getLoggerFromSomewhere();
const questionnaire = getQuestionnaireFromSomewhere();
const transformAndUpload = getTaskFromSomewhere();

const taskRunner = createTaskRunner({
    taskImplementations: {
        transformAndUpload
    },
    context: {
        questionnaire,
        logger
    }
});

const { task } = await taskRunner.run({
    id: "task1",
    type: "transformAndUpload",
    data: {
        questionnaire: "$.questionnaire",
        logger: "$.logger"
    }
});

task.result // submissionResponse
```

>     * I borrowed some code from the dataset service (I didn't extract this out into another helper to use in both under the assumption that the dataset service is going to be retired soon anyway from my understanding so that code won't be duplicated once that happens

Yeah, it'll get deprecated.

PR #366, opened 2023-07-03
Add metadata endpoints to openapi v2023-05-17

BarryPiccinni:


openapi/src/json-schemas/api/resources/metadata-resource.json
```diff
@@ -0,0 +1,48 @@
...
+                                "questionnaire-document-version",
+                                "questionnaire-id",
+                                "created",
+                                "modified",
+                                "submission-status",
+                                "expires"
+                            ],
+                            "properties": {
+                                "questionnaire-document-version": {
+                                    "enum": ["4.1.0"]
```
armoj, line 24: should this be abstracted out in its own schema to avoid this value deviating from the truth? 🤔 using the $ref syntax, and referencing a questionnaire-document-version schema with a static enum as a property.

The [latest ](https://github.com/CriminalInjuriesCompensationAuthority/q-templates-application/blob/master/src/template.js#L714)is `4.2.0`

openapi/src/openapi-src-v2.json
```diff
@@ -274,6 +280,176 @@
...
+                        "$ref": "#/components/parameters/apiVersion"
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "Success",
+                        "content": {
+                            "application/vnd.api+json": {
+                                "schema": {
+                                    "$ref": "json-schemas/api/v2_questionnaires_metadata/get/res/200.json"
```
armoj, line 419: Both metadata endpoints are using `json-schemas/api/v2_questionnaires_metadata/get/res/200.json` to define the shape. Is this right? One is a single resource and one is a collection (potentially of size:1), but they should be represented as such.

armoj (changes requested):

openapi/src/openapi-src-v2.json
BarryPiccinni: It was not right. I've updated this.

openapi/src/json-schemas/api/resources/metadata-resource.json
BarryPiccinni: I've updated this to 4.2.0.

I don't think it would make sense to abstract this out. We're not re-using the schema anywhere else, plus we'd still need to remember to keep it in sync. I don't think we'd gain much and it'd be inconsistent with all the other `enum` schema we've used.

openapi/src/json-schemas/api/resources/metadata-resource.json
```diff
@@ -0,0 +1,48 @@
...
+                                "questionnaire-document-version",
+                                "questionnaire-id",
+                                "created",
+                                "modified",
+                                "submission-status",
+                                "expires"
+                            ],
+                            "properties": {
+                                "questionnaire-document-version": {
+                                    "enum": ["4.2.0"]
```
armoj, line 24: After discussion, it has transpired that there is not requirement for this property at this time. Just remove it and it will reduce this complexity that we previously spoke about.

armoj (changes requested):

armoj (approved):
Looks good. Can you squash this in to 1 commit?

PR #367, opened 2023-07-05
Bp feature/save and return/dashboard code

BarryPiccinni:


questionnaire/metadata/metadata-service.test.js
```diff
@@ -0,0 +1,173 @@
...
+
+                const actual = await metadataService.getMetadataByQuestionnaire(
+                    validQuestionnaireId
+                );
+
+                expect(Array.isArray(actual.data)).toBe(true);
+                expect(actual.data).toMatchObject(expected);
+            });
+
+            it('Should error if the apiVersion is not "2023-05-17"', async () => {
```
armoj, line 144: Should we have a test for:
'Should error if the apiVersion is not "2023-05-17"'
'Should error if the apiVersion is undefined'

armoj (changes requested):

PR #380, opened 2023-07-28
Update aws sdk v3

jenniferbrown94:


BarryPiccinni (changes requested):
Line 16 of the `index.js` file (which I've commented on) is actually pulling your test coverage down under the threshold as well.

PR #385, opened 2023-08-08
Add new resubmission code for admin API

BarryPiccinni:


questionnaire/questionnaire-service.js
```diff
@@ -635,6 +635,23 @@ function createQuestionnaireService({
         );
     }
 
+    async function getQuestionnaireIdsBySubmissionStatus(status) {
+        return db.getQuestionnaireIdsBySubmissionStatus(status);
+    }
+
+    async function postFailedSubmissions() {
+        const questionnaireIds = await getQuestionnaireIdsBySubmissionStatus('FAILED');
+        const promises = questionnaireIds.map(async id => {
```
neil-stephen-mcgonigle, line 644: Is this a good name for this array 'promises'?
Something like appResubmissionState?
And curious what if anything is done with this data when it is returned?

questionnaire/questionnaire-service.js
```diff
@@ -635,6 +635,23 @@ function createQuestionnaireService({
...
+            try {
+                await startSubmission(id);
+                return {id, resubmitted: true};
+            } catch (err) {
+                return {id, resubmitted: false};
```
neil-stephen-mcgonigle, line 649: should we log this error?

questionnaire/questionnaire-service.js
```diff
@@ -635,6 +635,23 @@ function createQuestionnaireService({
...
+            }
+        });
+        return Promise.all(promises);
```
neil-stephen-mcgonigle, line 652: or leave the variable as is and replace with 
`return Promise.all(questionnaireIds.map(......)...`

questionnaire/questionnaire-service.test.js
```diff
@@ -844,4 +858,80 @@ describe('Questionnaire Service', () => {
             });
         });
     });
+
+    describe('DCS Admin API', () => {
```
neil-stephen-mcgonigle, line 862: Should we include a test for a failed submission when an error is thrown?

questionnaire/questionnaire-service.js
BarryPiccinni: This part of it was just a copy and paste from the existing `submission-routes.js`. This data is returned to the cronjob that pings this endpoint. It reports the result to Prometheus, I believe it actually just needs the length of the array rather than the actual object.

neil-stephen-mcgonigle (approved):

PR #391, opened 2023-08-16
Add questionnaire version comparison

BarryPiccinni:


questionnaire/questionnaire-service.js
```diff
@@ -448,6 +449,31 @@ function createQuestionnaireService({
     async function getProgressEntries(questionnaireId, query) {
         // 1 - get questionnaire instance
         const questionnaire = await getQuestionnaire(questionnaireId);
+        // 1.1 check questionnaire is a compatible version
+        const isCompatible = checkQuestionnaireVersion(questionnaire.version);
```
armoj, line 453: checkQuestionnaireVersion is not too descriptive of it's purpose here.

Have a function (util) with a sole purpose of checking if `isQuestionnaireVersionCompatible`. This util would then employ your `getInstalledModuleVersion` util and the `versionComparison` util.

````js
const isCompatible = isQuestionnaireVersionCompatible(questionnaire.version)// questionnaire instance version.

could be a case of a simple rename if I understand this fully.

**Review by armoj — CHANGES_REQUESTED** [review]

File: questionnaire/questionnaire-service.js [hunk 2]
```diff
@@ -448,6 +449,31 @@ function createQuestionnaireService({
...
         const questionnaire = await getQuestionnaire(questionnaireId);
+        // 1.1 check questionnaire is a compatible version
+        const isCompatible = isQuestionnaireCompatible(questionnaire.version);
+        // 1.2 if not, return 'incompatible questionnaire' schema
+        if (!isCompatible) {
+            return {
+                data: [
+                    {
+                        type: 'progress-entries',
+                        id: '0',
```
armoj, line 460: As discussed, change `id` to `incompatible`

questionnaire/utils/getInstalledModuleVersion/index.test.js
```diff
@@ -0,0 +1,60 @@
...
+                'no-version': undefined
+            }
+        }
+    };
+});
+
+const getVersion = require('.');
+
+describe('getInstalledModuleVersion', () => {
+    describe('No node_modules directory is found', () => {
```
armoj, line 22: strictly speaking, this is testing if the package-lock file is undefined.

questionnaire/utils/isQuestionnaireVersionCompatible/index.test.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
+
+beforeEach(() => {
+    jest.clearAllMocks();
+    jest.resetModules();
+});
+
+jest.doMock('semver/functions/diff', () => version => {
```
armoj, line 8: Don't mock things we don't own. Mock the template instance instead. Instantiate a template with a specific version in the json, and compare that against a specific question template version, again, that can be mocked (via the getInstalledModuleVersion function)

armoj (changes requested):

armoj (approved):

PR #398, opened 2023-08-24
[CDBT-187] Include description of where injury is on body in PDF

jenniferbrown94:
Update the themed json that is stored in S3 to include more details about injuries


sinclairs (approved):
LGTM

PR #403, opened 2023-09-07
Update resubmission function for new submission flow

BarryPiccinni:
Changes made: 
 - Moved resubmission function from `questionnaire-service.js` to new `submission-service.js`
 - Moved resubmission route from `routes.js` to `submission-routes.js`
 - Skipped a failing test in the generated test file that related to the old submission flow


questionnaire/submissions/submissions-routes.js
```diff
@@ -70,4 +70,19 @@ router
...
     });
 
+router.route('/resubmit-failed').post(permissions('admin'), async (req, res, next) => {
+    let response;
+    try {
+        const submissionService = createSubmissionService({
+            logger: req.log,
+            apiVersion: undefined,
+            ownerId: undefined
+        });
```
armoj, line 80: don't pass these in if they are not needed

armoj (changes requested):

armoj (approved):

PR #406, opened 2023-09-13
Rc save and return

armoj:
* Add questionnaire template version check
* Add "v2" submission tests
* Bump `q-template-application`

**Proposed version bump: `v8.1.0` (minor)**

### DIFF
--- a/questionnaire/utils/isQuestionnaireVersionCompatible/index.test.js
+++ b/questionnaire/utils/isQuestionnaireVersionCompatible/index.test.js
@@ -1,0 +1,44 @@
+'use strict';
+
+const baseVersion = '1.0.0';
+const majorVersionBump = '2.0.0';
+const minorVersionBump = '1.1.0';
+const patchVersionBump = '1.0.1';
+
+beforeEach(() => {
+    jest.clearAllMocks();
+    jest.resetModules();
+});
+
+jest.doMock('../getInstalledModuleVersion', () => () => baseVersion);
+const isVersionValid = require('.');
+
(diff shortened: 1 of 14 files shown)


questionnaire/utils/isQuestionnaireVersionCompatible/index.test.js
```diff
@@ -0,0 +1,44 @@
...
+
+describe('VersionComparison', () => {
+    describe("There's a major semver difference", () => {
+        it('Should return false', () => {
+            const actual = isVersionValid(majorVersionBump);
+            expect(actual).toBe(false);
+        });
+    });
+
+    describe("There's a minor semver difference", () => {
```
neil-stephen-mcgonigle, line 24: I am confused by this test? 
Why are we moving away from SemVer convention?
I get this does not _need_ to be the same but curious as to why it isn't.

neil-stephen-mcgonigle (approved):

questionnaire/utils/isQuestionnaireVersionCompatible/index.test.js
BarryPiccinni: @neil-stephen-mcgonigle We're still using the Semver pattern and the same rules for versioning as before. Ideally, we'd only want a major template change to be "incompatible", but we've got some coupling between CW and DCS (eg the UISchema example Adrian mentioned) that means we can't guarantee this. So for now we've decided to err on the side of caution and say any Major or Minor changes will be considered incompatible and applicants will need to start a new questionnaire.

PR #422, opened 2023-11-30
Dcs deploy

neil-stephen-mcgonigle:
technically a breaking change
but we are going to go ahead and patch and deal with any saved applications that have characters > 50 which are then submitted. (They will appear on the tempus DLQ).
If someone goes back to the question with > 50 chars they will encounter a validation error.


neil-stephen-mcgonigle:
wrong branch

PR #427, opened 2023-12-01
Fix/task runner logging

neil-stephen-mcgonigle:
WIP: for consideration


questionnaire/questionnaire/utils/taskRunner/tasks/sequential/index.js
```diff
@@ -1,24 +1,33 @@
...
-async function runTasksSequentially(taskDefinitions, run) {
-    const taskStates = [];
+class SequentialTaskError extends Error {
+    constructor(message, cause, taskIndex) {
+        super(message);
+        this.name = 'SequentialTaskError';
+        this.cause = cause;
+        this.taskIndex = taskIndex;
+    }
+}
```
sinclairs, line 10: The extended errors approach looks good. I like that the newish `cause` option has good interoperability with other modules 👌 e.g. the pino serializers. That should allow us to eventually deprecate VError;

I reckon we should go with the guidance of moving this functionality out to a central error location (if we had that, the VError refactoring would have been easy!).

I've thrown (PUN) up a [gist](https://gist.github.com/sinclairs/8ded99d0f8b6b63eb49d743891d75c04) with some additional ideas. Would be good to go through this and see how we want to take it forward e.g. the guidance suggests not creating an error per thing, but I could see value in being able to validate specific error args 🤔

[An example of how we discriminate between errors](https://github.com/CriminalInjuriesCompensationAuthority/data-capture-service/blob/5e36eb422c63ff55024b39f264a43b2350651da8/middleware/error-handler/index.js#L96)

sinclairs (changes requested):
Given the the logging and refactoring concerns in this PR are based on a revised error approach, I've just reviewed that aspect. Be good to catch up and get some thoughts 👍

questionnaire/questionnaire/utils/taskRunner/tasks/sequential/index.js
neil-stephen-mcgonigle: All good. I would suggest we get this out and create a new ticket off of the back of this to centralise the logging and select how we want to handle the errors in terms of differentiation. 
That way we can get the value of logging any submission errors we do encounter until the central logging is in place?
Here's an other decent [article](https://www.toptal.com/nodejs/node-js-error-handling)

neil-stephen-mcgonigle: @sinclairs any thoughts on my comment?

PR #430, opened 2023-12-21
Feature/central error logging

neil-stephen-mcgonigle:
Talked over with Stephen this is good to go after review.

### DIFF
--- a/questionnaire/submissions/submissions-service.test.js
+++ b/questionnaire/submissions/submissions-service.test.js
@@ -41,7 +41,7 @@ jest.doMock('../questionnaire-dal.js', () =>
                 };
             }
 
-            // submittable but unsuccessful due to failing task
+            // submittable but unsuccessful due to failing task, default retries
             if (questionnaireId === '93a1d3ab-56a6-4554-8acf-cdeb088d511d') {
                 return {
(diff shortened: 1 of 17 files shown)


neil-stephen-mcgonigle:
I will add another test for the [error-handler](https://github.com/CriminalInjuriesCompensationAuthority/data-capture-service/blob/57f1731c9ad4206ba6858b6e7dc0d18ebf09c994/middleware/error-handler/index.js#L16)

questionnaire/submissions/submissions-service.test.js
```diff
@@ -41,7 +41,7 @@ jest.doMock('../questionnaire-dal.js', () =>
                 };
             }
 
-            // submittable but unsuccessful due to failing task
+            // submittable but unsuccessful due to failing task, default retires
```
armoj, line 44: do you mean "default retries" here?

armoj (changes requested):

questionnaire/submissions/submissions-service.test.js
neil-stephen-mcgonigle: yup

armoj (approved):

PR #431, opened 2023-12-29
CDBT-308

msenkiw:
https://dsdmoj.atlassian.net/browse/CDBT-308

Currently, we send files to a generic application_json folder. This PR updates it so that the files are sent to a folder which is called the application CRN.


sinclairs (approved):
Spoke with Jack. This has been tested and is good to go. It's a backwards compatible change.

PR #432, opened 2024-01-08
Bp refactor/save and return/remove redundant code

BarryPiccinni:


neil-stephen-mcgonigle (commented):
@BarryPiccinni just to let you know, I am looking at this.
One thing I have noticed is that there are still a lot of unused methods(only used by tests) within the questionnaire-dal.js Is this something you are aware of and plan to address?
I have been looking into tools to identify these issues but with no success so far.
For example [knip](https://knip.dev/)
image

BarryPiccinni:
> @BarryPiccinni just to let you know, I am looking at this. One thing I have noticed is that there are still a lot of unused methods(only used by tests) within the questionnaire-dal.js Is this something you are aware of and plan to address? I have been looking into tools to identify these issues but with no success so far. For example [knip](https://knip.dev/) image

I've not edited the DAL too much, so yeah, there will be a lot of functions that are now unused. The problem is the new submission process is still calling some of the "V1" version of these queries. Ideally I'd like everything to be consistent and use the new versions but that's quite a big refactor in it's own right, reckon it commands it's own ticket!

armoj (approved):

PR #436, opened 2024-01-18
Add 'origin' to API and send with answers

BarryPiccinni:


questionnaire/routes.test.js
armoj, line 1: Can we have tests for:

- 'W'
- 'T'
- something invalid
- not including `origin` in the object at all

questionnaire/questionnaire/utils/taskRunner/tasks/transformAndUpload/index.js
```diff
@@ -135,7 +135,8 @@ function transformQuestionnaire(questionnaire) {
     return {
         meta: {
             caseReference: questionnaire.getAnswers().system['case-reference'],
-            funeralReference: questionnaire.getAnswers().system['secondary-reference']
+            funeralReference: questionnaire.getAnswers().system['secondary-reference'],
+            origin: questionnaire.getAnswers().owner.origin
```
armoj, line 139: should this be `ownerOrigin` to make it clear what it means, an keep it inline with the `owner.origin` object property in terms of name/namespace.

armoj (changes requested):

questionnaire/questionnaire/utils/taskRunner/tasks/transformAndUpload/index.js
BarryPiccinni: I don't think that's necessary. We've currently already got `owner.id` and `owner.isAuthenticated` we've already set the precedent for naming things this way.

PR #439, opened 2024-01-24
Add origin answer to API

BarryPiccinni:


questionnaire/questionnaire/utils/taskRunner/tasks/transformAndUpload/index.js
```diff
@@ -132,14 +132,21 @@ function transformQuestionnaire(questionnaire) {
...
-            funeralReference: questionnaire.getAnswers().system['secondary-reference']
+            caseReference: answers.system['case-reference'],
+            funeralReference: answers.system['secondary-reference']
         },
         themes: themeContent,
         declaration: getDeclaration(questionnaire)
     };
+
+    if (answers.origin) {
+        transformedQuestionnaire.meta.channel = answers.origin.channel;
```
armoj, line 146: Should we keep a reference to `origin` here? 🤔

armoj (changes requested):

questionnaire/questionnaire/utils/taskRunner/tasks/transformAndUpload/index.js
BarryPiccinni: The transformed questionnaire is a flattened view, I was naming this on the basis that the bit of data we're interested in is the "channel" value, in the same way we lose reference to "system" when mapping "caseReference" above.

armoj: makes absolute sense! 👍

armoj (approved):

PR #449, opened 2024-05-13
Move UI schema properties

tjbburton:
- Install the new templates which include UI schema information
- Sort the CYA page summary structure before returning it, rather than sorting it in q-transformer. This prevents differences in ordering between the PDF and CYA page

### DIFF
--- a/questionnaire/questionnaire/questionnaire.js
+++ b/questionnaire/questionnaire/questionnaire.js
@@ -10,6 +10,7 @@ defaults.getValueContextualiser = require('./utils/getValueContextualiser');
 defaults.deepClone = require('./utils/deepCloneJsonDerivedObject');
 defaults.getJsonExpressionEvaluator = require('./utils/getJsonExpressionEvaluator');
 defaults.qExpression = require('q-expressions');
+defaults.sortThemedAnswers = require('./utils/sortThemedAnswers');
 
 function createQuestionnaire({
     questionnaireDefinition,
@@ -21,7 +22,8 @@ function createQuestionnaire({
     getValueContextualiser = defaults.getValueContextualiser,
     deepClone = defaults.deepClone,
     getJsonExpressionEvaluator = defaults.getJsonExpressionEvaluator,
-    qExpression = defaults.qExpression
+    qExpression = defaults.qExpression,
+    sortThemedAnswers = defaults.sortThemedAnswers
 }) {
     function getId() {
         return questionnaireDefinition.id;
@@ -250,6 +252,15 @@ function createQuestionnaire({
 
         if (sectionDefinitionVars !== undefined && allowSummary === true) {
             const resolvedVars = getResolvedVars(sectionId, sectionDefinitionVars);
+            if (resolvedVars.summary) {
+                if (sectionDefinition.schema.options) {
+                    const sortingInstructions = sectionDefinition.schema.options.ordering;
+                    resolvedVars.summary = sortThemedAnswers(
+                        resolvedVars.summary,
+                        sortingInstructions
+                    );
+                }
+            }
             const valueVarReplacer = getValueVarReplacer(resolvedVars);
 
             orderedValueTransformers.push(valueVarReplacer);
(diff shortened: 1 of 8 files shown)


questionnaire/questionnaire/questionnaire.js
```diff
@@ -190,6 +190,40 @@ function createQuestionnaire({
...
+        const currentQuestionId = values[valueIndex].id;
+        const previousQuestionId = values[valueIndex - 1].id;
+        const dependentIds = getDependentQuestionIds(currentQuestionId, sortingInstructions);
+        return dependentIds.includes(previousQuestionId);
+    }
+
+    function sortThemedAnswers(themeContent, sortingInstructions) {
+        themeContent.forEach(theme => {
+            if (theme.type === 'theme') {
+                const values = theme.values ? theme.values : [];
```
neil-stephen-mcgonigle, line 211: Avoid Ternary Operator: Instead of using ternary operator, you can use the logical AND (&&) operator to assign a default value.
const values = theme.values || [];

questionnaire/questionnaire/questionnaire.js
```diff
@@ -190,6 +190,40 @@ function createQuestionnaire({
...
+        return questionOrdering || [];
+    }
+
+    function dependentsInIncorrectOrder(values, valueIndex, sortingInstructions) {
+        if (valueIndex === 0) {
+            return false;
+        }
+        const currentQuestionId = values[valueIndex].id;
+        const previousQuestionId = values[valueIndex - 1].id;
+        const dependentIds = getDependentQuestionIds(currentQuestionId, sortingInstructions);
```
neil-stephen-mcgonigle, line 204: Performance Improvement: If dependentsInIncorrectOrder is a bottleneck due to a large number of dependentIds and previousQuestionIds, consider converting dependentIds to a Set for faster lookup.
const dependentIds = new Set(getDependentQuestionIds(currentQuestionId, sortingInstructions));
    return dependentIds.has(previousQuestionId);

questionnaire/questionnaire/questionnaire.js
```diff
@@ -190,6 +190,40 @@ function createQuestionnaire({
...
+    function getDependentQuestionIds(questionId, sortingInstructions) {
+        const questionOrdering = sortingInstructions[questionId];
+        return questionOrdering || [];
+    }
+
+    function dependentsInIncorrectOrder(values, valueIndex, sortingInstructions) {
+        if (valueIndex === 0) {
+            return false;
+        }
+        const currentQuestionId = values[valueIndex].id;
```
neil-stephen-mcgonigle, line 202: Use Destructuring: Utilize destructuring where applicable to simplify code and improve readability.
const { id: currentQuestionId } = values[valueIndex];
const { id: previousQuestionId } = values[valueIndex - 1];

questionnaire/questionnaire/questionnaire.js
tjbburton: I don't think set conversion is necessary here as we only ever have a few dependentIds in the list (the most is for the physical injury question, which only has 33)

neil-stephen-mcgonigle: agreed, just putting it out as an example and somethign to think about in the future

questionnaire/questionnaire/questionnaire.js
tjbburton: I think that destructuring is actually less readable here, as we are not assigning the Id property to a variable of the same name

neil-stephen-mcgonigle: { id: currentQuestionId }: This is an object destructuring syntax. It means that from the object retrieved from values[valueIndex], we want to extract the property id and assign it to a variable named currentQuestionId.
Happy to leave as is though as arguably this is less readable

neil-stephen-mcgonigle:
oh and should we refactor this code out to a utility?

neil-stephen-mcgonigle (approved):

PR #453, opened 2024-06-17
Implement the task list-supported template

armoj:


questionnaire/task-list/task-list-service.js
```diff
@@ -0,0 +1,91 @@
...
+    function isTaskListSchema({sectionSchema, sectionId} = {}) {
+        if (sectionSchema) {
+            const pageType = sectionSchema?.meta?.pageType;
+            if (pageType === 'task-list') {
+                return true;
+            }
+        }
+
+        // TODO: pass in the sectionSchema, not the sectionId. OR
+        // TODO: get schema from sectionId and interrogate the shape instead of the IDs value.
```
BarryPiccinni, line 25: Another reason to improve the progress entries function

questionnaire/task-list/task-list-service.js
```diff
@@ -0,0 +1,91 @@
...
+        // the lower the index, the higher the precedent.
+        if (
+            taskStatusPrecedence.indexOf(taskCompletionStatus) <
+            taskStatusPrecedence.indexOf(taskApplicabilityStatus)
+        ) {
+            return taskCompletionStatus;
+        }
+
+        return taskApplicabilityStatus;
+    }
```
BarryPiccinni, line 63: I understand what you've done here, but I think it should be:

```
const status = taskApplicabilityStatus = 'applicable' ? taskCompletionStatus : taskApplicabilityStatus
```
It's not a perfect solution but I think it's just far easier to read and understand than these precedence arrays.

BarryPiccinni (changes requested):

PR #473, opened 2024-10-07
Initial action functionality

tjbburton:


tjbburton:
The 'get whether an action is onComplete or onCreate from template' commit allows us to distinguish between actions to run on creation and submission, for example confirmation emails vs notification emails. The alternative to specifically definining them would be to infer submission vs creation using conditions, for example if the first question in a template has been answered then it is submission.

PR #521, opened 2026-03-02
update test fixtures survey links

szabinakov:
Updated survey links to smart survey in test fixtures


BarryPiccinni (changes requested):
Changes are fine. Could you tidy up the commit messages tho? Unless you think each individual commit has value, generally we squash whatever we've done into one commit, then follow up with a version commit, that way you end up with one commit containing all your work, followed by a second commit announcing the new version. This makes it clear where the releases are in the commit history. Also, it's useful to have the q-templates version you've updated to in the commit message. Makes your change easier to find. So rather than "Bumped to latest version of q-templates" make it something like "Installed q-templates v1.2.3"

duncanabraham (approved):
All good 👍 With reference to Barry's comment, have a look here: https://www.conventionalcommits.org/en/v1.0.0/

BarryPiccinni (approved):
LGTM

PR #542, opened 2026-04-21
Add owner data to metadata

BarryPiccinni:


questionnaire/questionnaire-service.js
tjbburton, line 633: Probably worth changing how this function returns 'personalisation data' as well as we are moving away from that block for names anyway

tjbburton (changes requested):

PR #544, opened 2026-04-27
snyk setup

szabinakov:
Two different scan set up:

- `snyk_app_scan  `: scans the app's dependencies (package.json) for known CVEs. Runs early in the pipeline before building the Docker image.
- The Snyk container scan in the build job: scans the built Docker image including the OS, Node runtime, and npm itself.

Two different monitoring set up:

- `snyk_monitor` : monitoring the app dependencies (only prod deployment branch)
- `snyk_container_monitor`: monitoring the built img (only prod deployment branch)

This PR also includes an upgrade to node `24.15.0` to fix some flagged vulnerabilities, and an upgrade to npm `11.13.0` in the Dockerfile due to the base img (node 24.15.0) using older vulnerable npm version.

npm in package.json is upgraded to `11.12.1` which is the supported version in node `24.15.0`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,9 +13,34 @@ executors:
       # Image tool container
       - image: ministryofjustice/cloud-platform-tools
 jobs:
+  snyk_app_scan:
+    docker:
+      - image: node:24.15.0-trixie-slim
+    resource_class: large
+    steps:
+      - checkout
+      - restore_cache:
+          key: v1-npm-{{checksum "package-lock.json" }}
+      - run:
+          name: Install CA certificates
+          command: |
+            apt-get update
+            apt-get install -y ca-certificates
+      - run:
+          name: Install Dependencies
+          command: npm ci
+      - run:
+          name: Install Snyk
+          command: npm install -g snyk@1.1304.0
+      - run:
+          name: Snyk test
+          command: snyk test --severity-threshold=high
+      - run:
+          name: Snyk monitor
+          command: snyk monitor
   build-for-test:
     docker:
-      - image: node:24.13.0-trixie-slim
+      - image: node:24.15.0-trixie-slim
     resource_class: large
     steps:
       - checkout
@@ -37,7 +62,7 @@ jobs:
             - ./node_modules
   test:
     docker:
-      - image: node:24.13.0-trixie-slim
+      - image: node:24.15.0-trixie-slim
     steps:
       - attach_workspace:
           at: ./
@@ -48,7 +73,7 @@ jobs:
           command: npx --no-install jest --ci --runInBand --bail --silent --coverage --projects jest.config.js
   lint:
     docker:
-      - image: node:24.13.0-trixie-slim
+      - image: node:24.15.0-trixie-slim
     steps:
       - attach_workspace:
           at: ./
@@ -66,6 +91,22 @@ jobs:
           name: Build docker image
           command: |
             docker build --target production --no-cache -t cica/cica-repo-dev .
+
+      - run:
+          name: Install Snyk CLI
+          command: |
+            apk add --update-cache curl
+            mkdir -p /usr/local/bin
+            curl -sL https://github.com/snyk/cli/releases/download/v1.1304.0/snyk-alpine -o /usr/local/bin/snyk
+            chmod +x /usr/local/bin/snyk
+            /usr/local/bin/snyk --version
+
+      - run:
+          name: Snyk container scan
+          command: |
+            /usr/local/bin/snyk container test cica/cica-repo-dev --file=Dockerfile --severity-threshold=high 
+            /usr/local/bin/snyk container monitor cica/cica-repo-dev
+
       - run:
           name: Archive Docker Image
           command: |
@@ -218,10 +259,14 @@ workflows:
       - lint:
           requires:
             - build-for-test
+      - snyk_app_scan:
+          requires:
+            - build-for-test
       - build:
           requires:
             - test
             - lint
+            - snyk_app_scan
       - publish_latest:
           requires:
             - build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,7 +20,7 @@ jobs:
     steps:
       - checkout
       - restore_cache:
-          key: v1-npm-{{checksum "package-lock.json" }}
+          key: v1-npm-{{ checksum "package.json" }}-{{checksum "package-lock.json" }}
       - run:
           name: Install CA certificates
           command: |
@@ -35,9 +35,48 @@ jobs:
       - run:
           name: Snyk test
           command: snyk test --severity-threshold=high
+  snyk_monitor:
+    docker:
+      - image: node:24.15.0-trixie-slim
+    resource_class: large
+    steps:
+      - checkout
+      - restore_cache:
+          key: v1-npm-{{ checksum "package.json" }}-{{checksum "package-lock.json" }}
+      - run:
+          name: Install Dependencies
+          command: npm ci
+      - run:
+          name: Install Snyk
+          command: npm install -g snyk@1.1304.0
+      - run:
+          name: Install CA certificates
+          command: |
+            apt-get update
+            apt-get install -y ca-certificates
       - run:
           name: Snyk monitor
           command: snyk monitor
+  snyk_container_monitor:
+    executor: docker-publisher
+    steps:
+      - attach_workspace:
+          at: /tmp/workspace
+      - setup_remote_docker
+      - run:
+          name: Load archived Docker image
+          command: docker load -i /tmp/workspace/image.tar
+      - run:
+          name: Install Snyk CLI
+          command: |
+            apk add --update-cache curl
+            mkdir -p /usr/local/bin
+            curl -sL https://github.com/snyk/cli/releases/download/v1.1304.0/snyk-alpine -o /usr/local/bin/snyk
+            chmod +x /usr/local/bin/snyk
+      - run:
+          name: Snyk container monitor
+          command: |
+            /usr/local/bin/snyk container monitor cica/cica-repo-dev
   build-for-test:
     docker:
       - image: node:24.15.0-trixie-slim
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -218,10 +291,28 @@ workflows:
...
+      - snyk_app_scan:
+          requires:
+            - build-for-test
+      - snyk_monitor:
+          requires:
+            - build-for-test
+          filters:
+            branches:
+              only:
+                - cw-deploy
```
neil-stephen-mcgonigle: cw-deploy is this correct?

.circleci/config.yml
```diff
@@ -218,10 +291,28 @@ workflows:
...
+      - snyk_container_monitor:
+          requires:
+            - push_image
+          filters:
+            branches:
+              only:
+                - cw-deploy
```
neil-stephen-mcgonigle: cw-deploy again

.circleci/config.yml
neil-stephen-mcgonigle: looks like you fixed this

.circleci/config.yml
neil-stephen-mcgonigle: and this

.circleci/config.yml
```diff
@@ -218,10 +291,28 @@ workflows:
...
+      - snyk_app_scan:
+          requires:
+            - build-for-test
+      - snyk_monitor:
+          requires:
+            - build-for-test
+          filters:
+            branches:
+              only:
+                - dcs-deploy
```
neil-stephen-mcgonigle, line 309: Is it this branch or master we want to monitor?
This branch, [dcs-deploy](https://github.com/ministryofjustice/cica-apply-data-capture-service/tree/dcs-deploy), is [41 commits ahead of](https://github.com/ministryofjustice/cica-apply-data-capture-service/compare/master...dcs-deploy) master.

neil-stephen-mcgonigle:
The readme is out of date?
https://github.com/ministryofjustice/cica-apply-data-capture-service/blob/7da9e577d53bd75e6bb21a28a6c19d5441645ae0/README.md?plain=1#L19

.circleci/config.yml
neil-stephen-mcgonigle: Or both?

szabinakov: I think deploy is enough, should be the same as master

szabinakov:
> The readme is out of date?
> 
> https://github.com/ministryofjustice/cica-apply-data-capture-service/blob/7da9e577d53bd75e6bb21a28a6c19d5441645ae0/README.md?plain=1#L19

It is updated in this PR

.circleci/config.yml
neil-stephen-mcgonigle: is it? https://github.com/ministryofjustice/cica-apply-data-capture-service/compare/master...dcs-deploy

neil-stephen-mcgonigle: https://app.circleci.com/pipelines/github/ministryofjustice/cica-apply-data-capture-service?branch=dcs-deploy

The current release and production branch is master which was deployed 17 days ago.

neil-stephen-mcgonigle: so adding a second scan right before prod deploy may be a sensible hardening step for long-lived approval queues, which is the current state.

neil-stephen-mcgonigle: Master could currently have vulnerabilities that have not been picked up and should be addressed/monitored?

Dockerfile
```diff
@@ -1,5 +1,8 @@
 # lets start from an image that already has nodejs installed
-FROM node:24.13.0-trixie-slim AS base
+FROM node:24.15.0-trixie-slim AS base
```
neil-stephen-mcgonigle, line 2: security hardening recommendation:
Pin production-relevant images by digest, using tag plus digest together for readability and intent:

`FROM node:24.15.0-trixie-slim@sha256:... AS base`

Do the same for CircleCI job images in config.yml, not just Dockerfile, as CI runner images can drift too.

Why this is worth it:

- Prevents silent upstream tag drift.
- Improves supply-chain integrity and incident forensics.
- Makes builds deterministic across time.

Tradeoff:

We must rotate digests intentionally as patch releases arrive.
Best handled with automation (Renovate/Dependabot Docker digest updates + scheduled refresh PRs). 
And we could look at this later as in create a new ticket to address adding and potentially automatiing image digest updates.

.circleci/config.yml
neil-stephen-mcgonigle: So monitoring master is probably something we want to set up independently from this workflow, should be easier when migrating to gitbub actions. But adding a pre-deployment scan may be an idea.

szabinakov: Fair point, I’ll change this to monitor master, since that’s our production branch.
I’d prefer not to add an additional scan later in the pipeline, as it starts to feel duplicative and adds complexity. Ideally vulnerabilities should be caught earlier, not at deploy time.
If things are regularly sitting in dev/prod approval for days, I think the better fix is to understand why that’s happening and work on speeding that flow up, rather than compensating with extra scans.

.circleci/config.yml
```diff
@@ -66,6 +124,21 @@ jobs:
           name: Build docker image
           command: |
             docker build --target production --no-cache -t cica/cica-repo-dev .
+
+      - run:
+          name: Install Snyk CLI
```
neil-stephen-mcgonigle, line 135: Did you consider or attempt to use the[ snyk circle orb](https://circleci.com/developer/orbs/orb/snyk/snyk)? 
It is being used elsewhere in the organisation.
[Snyk orb example](https://github.com/ministryofjustice/hmpps-circleci-orb/blob/a46836971b2e65326756a0bea772f757cade877f/src/%40orb.yml#L16)

Then this could be changed to something along these lines.. (as could the other run commands)

orbs:
  aws-cli: circleci/aws-cli@4.0.0
  snyk: snyk/snyk@2.3.0

      - snyk/scan:
          cli-version: 1.1304.0
          docker-image-name: cica/cica-repo-dev:latest
          monitor-on-build: false
          os: alpine
          severity-threshold: high
          target-file: Dockerfile


Otherwise consider hardening to

- run:
    name: Install Snyk CLI (verified)
    command: |
      set -euo pipefail
      apk add --no-cache curl
      SNYK_VERSION="1.1304.0"
      SNYK_URL="https://github.com/snyk/cli/releases/download/v${SNYK_VERSION}/snyk-alpine"
      SNYK_SHA256="<expected_sha256_here>"

      curl -fsSL --retry 3 --retry-delay 2 "$SNYK_URL" -o /usr/local/bin/snyk
      echo "${SNYK_SHA256}  /usr/local/bin/snyk" | sha256sum -c -
      chmod 0755 /usr/local/bin/snyk
      /usr/local/bin/snyk --version

as there are high risk concerns:

- Unverified binary download
The job downloads an executable from the internet and runs it without integrity verification in config.yml:68. If the artifact is tampered with, CI would execute it.
- Weak curl failure behavior
Using only -sL in config.yml:68 can hide errors. You want fail-fast semantics so bad HTTP responses do not silently produce bad files.
- Root execution path
The binary is written to /usr/local/bin and executed as root in this executor path config.yml:66-69. In ephemeral CI this is common, but it increases blast radius if supply chain is compromised.


Using the orb would be my recommendation thus following organisational patterns.

.circleci/config.yml
neil-stephen-mcgonigle: But when deploy time may happen weeks later, if the dcs-deploy currently sitting at awaiting approval is then approved and deployed the scan is out of date. This may be considered an edge case, I am not sure whether this will happen or happens often, at least monitoring master would pick it up though after it had been deployed. 

There is an example of a [cronjob](https://github.com/ministryofjustice/hmpps-probation-in-court-utils/blob/56810f9cf20d3746eea1a28e0b9c95c9ee8d9293/.circleci/config.yml#L97) here for montioring main/master if that helps?

.circleci/config.yml
szabinakov: I did look at the Snyk orb, but unfortunately it isn’t compatible with the ministryofjustice/cloud-platform-tools image we’re required to use for this job.
I understand the concerns around downloading and executing a binary in CI. In this case the version is pinned, the environment is ephemeral, and the execution scope is limited to the job, so I’m comfortable with the residual risk.
Given those constraints, I think this is a proportionate implementation and I’d prefer to avoid additional complexity, particularly as this configuration is temporary and will be replaced when we move to GitHub Actions in the coming weeks.

.circleci/config.yml
szabinakov: The deployments stuck in dev/uat/prod for days and weeks is not really the usual. The last couple weeks were interrupted with accessibility testing / pen testing etc. 
For now the goal is to have something in the pipeline that replaces trivy, until we move to GA. The idea is to see how this monitoring/scanning set up behaves, and after a couple weeks we can revisit if we feel like we would benefit from tighten or loosen either. As I mentioned in the slack message when sending the PRs it would be worth to consider to merge dependabot and snyk too. As I am not sure if we have the manpower to be on top of all this constantly.

Dockerfile
szabinakov: I think this is a good point as well, probably worth revisiting when all flow reworked in GA

.circleci/config.yml
neil-stephen-mcgonigle: I understand the goal, I am pointing out things that I am picking up on whilst reviewing this pull request. Feel free to add to the slack chat.or note, for upcoming discussions.

.circleci/config.yml
neil-stephen-mcgonigle: I wonder how the other teams are working with the snyk orb and  ministryofjustice/cloud-platform-tools? As long as you have looked into this and you are aware of the security concerns I will approve given that these projects shall be moving to github actions in the near term, when these concerns can, hopefully, be addressed and improvvements made.

neil-stephen-mcgonigle (dismissed):

.circleci/config.yml
szabinakov: They dont, I went through the entire ministriy of justice organisation and looked at repos where snyk orb is used, they don't scan the images built with the cloud-platform-tool.

[Here ](https://github.com/ministryofjustice/laa-crime-hardship/blob/42710b209f9165bd2d50136fe1ced011928a412f/.circleci/config.yml#L4)is an example.

neil-stephen-mcgonigle: That's good to know, I haven't looked at it in depth I am not familiar with how these [executors](https://github.com/ministryofjustice/laa-crime-hardship/blob/42710b209f9165bd2d50136fe1ced011928a412f/.circleci/config.yml#L24-L25) work.

neil-stephen-mcgonigle (approved):

neil-stephen-mcgonigle (approved):

PR #548, opened 2026-06-03
user trace in logs

szabinakov:
- add questionnaireId and ownerId to each dcs log
- send the questionnaireId to the notify and the application service

### DIFF
--- a/app.js
+++ b/app.js
@@ -11,6 +11,7 @@ const docsRouter = require('./docs/routes');
 const questionnaireRouter = require('./questionnaire/routes');
 
 const app = express();
+
 const logger = pino({
     level: process.env.DCS_LOG_LEVEL,
     redact: {
--- a/middleware/request-context/index.js
+++ b/middleware/request-context/index.js
@@ -1,0 +1,47 @@
+'use strict';
+
+const crypto = require('crypto');
+
+function resolveQuestionnaireId(req) {
+    return req.params?.questionnaireId || req.body?.data?.attributes?.questionnaireId || undefined;
+}
+
+function baseContextMiddleware(req, res, next) {
+    const ownerId = req.get('On-Behalf-Of');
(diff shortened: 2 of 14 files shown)


app.js
```diff
@@ -41,7 +48,11 @@ const logger = pino({
         }
 
         return 'info';
-    }
+    },
+    customProps: req => ({
+        questionnaireId: extractQuestionnaireIdFromUrl(req.url),
+        ownerId: req.headers['on-behalf-of']
+    })
```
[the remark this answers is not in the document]
szabinakov: The web-app actually sends the owner id as on-behalf-of on creation:
https://github.com/ministryofjustice/cica-apply-web/blob/e7ead1dc11b97a784dbd9dfa88ab2b1552c200ef/questionnaire/questionnaire-service.js#L15-L16

app.js
```diff
@@ -11,6 +11,13 @@ const docsRouter = require('./docs/routes');
 const questionnaireRouter = require('./questionnaire/routes');
 
 const app = express();
+
+// Extracts the questionnaireId UUID from a DCS API URL path.
+function extractQuestionnaireIdFromUrl(url) {
+    const match = url && url.match(/\/questionnaires\/([^/?]+)/);
+    return match ? match[1] : undefined;
+}
```
BarryPiccinni: Is this the only way to get the questionnaireId at this point? It's a little bit brittle.

szabinakov: It only appears in the request's url. 😞

armoj: is it not in the session at the point where you _could_ extract it out?

szabinakov: The logger has to run before the route handlers so that req.log is available throughout the application. At that point in the flow, express hasnt processed the route yet so req.param is empty

armoj (approved):

PR #549, opened 2026-06-18
Snyk fixes

szabinakov:
- Updated snyk ignore (removed and added)
- upgraded fixed npm version in docker image to v17, to fix some vulnerabilities

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 FROM node:24.15.0-trixie-slim AS base
 
 USER root
-RUN npm install -g npm@11.13.0
+RUN npm install -g npm@11.17.0
 
 RUN groupadd -g 1014 dc_user \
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -53,6 +53,7 @@ FROM node:24.15.0-trixie AS dev
 
 RUN groupadd -g 1014 dc_user \
     && useradd -rm -d /usr/src/app -u 1015 -g dc_user dc_user
+RUN npm install -g npm@11.17.0
```
tjbburton, line 56: Do we need this second npm install?

szabinakov: Well, things a bit funky here. Locally we use(d) a different img and npm version when in prod. So if you run snyk scanning locally you are not getting the same results as in prod.
I added the install here too so we install the same fix npm version in the local run too. 
(the base img still differ tho, so not perfect, but one step closer to parity)

tjbburton (approved):

